PR #1, opened 2020-01-31
RFC00000: First version of an RFC process, based on Aries and Rust

axelsimon:
This modifies/adds four files:
- README, which contains the general description of the RFC process
- contributing, which explains the process more pragmatically
- a template, for submitters to fill
 - a picture, describing the process and used in the README

This process is heavily taken from Hyperledger Aries and the Rust language (leaning more towards Aries, which is itself based on the Rust process), but adapted to our needs and the current level of maturity of Enarx.

I expect this to create some discussion, so please do comment and suggest changes.
My main fear is that this is __too__ involved a process.


contributing.md
```diff
@@ -0,0 +1,106 @@
+# Enarx RFCs
+
+## Contributing
+
+### Do you need an RFC?
+
+Use an RFC to advocate substantial changes to the Aries ecosystem, where
```
lkatalin: I think we want it to say "Enarx" here. : )

contributing.md
```diff
@@ -0,0 +1,106 @@
...
+    poorly received. You can add supporting artifacts, such as diagrams and sample
+    data, in the RFC's folder.
+  - Consider how the RFC should be [tagged](/tags.md).
+  - Assign a number to your RFC. Get the number by inspecting open and closed PRs against
+    this repo to figure out what the next PR number will be). Rename your folder from
+    `<your folder name>` to `<your 4-digit number>-<your folder name>`. At the
+    top of your README.md, modify the title so it is in the form: `<your 4-digit
+    number>: Friendly Version of Your Title`. Commit your changes.
+  - Commit the updated version of /index.md and push your changes.
+  - Submit a pull request.
```
lkatalin: If anything, I think this is where people would get lost or do something incorrectly; wondering if there is any way to make this more streamlined / intuitive and not require so many instructions (although there may not be). At least we can always request changes on a PR.

lkatalin: Related thought: is it feasible to have some type of CI tests for this whole process to make sure people did do it correctly?

0000-template.md
```diff
@@ -0,0 +1,101 @@
...
+Discuss prior art, both the good and the bad, in relation to this proposal.
+A few examples of what this can include are:
+
+- Does this feature exist in other TEE-based systems and what experience have
+their community had?
+- For other teams: What lessons can we learn from other attempts?
+- Papers: Are there any published papers or great posts that discuss this?
+If you have some relevant papers to refer to, this can serve as a more detailed
+theoretical background.
+
```
lkatalin: This is a useful section.

lkatalin (commented):
I think the template is useful and the write-up is very thorough and seemed pretty clear. The set of instructions about making a PR and choosing your numbering scheme, etc., is somewhat complex and might serve as a deterrent. Does the Rust RFC also use this same PR-based method? I haven't gone through that process, so if it's similar then this probably works as well.

contributing.md
MikeCamel: This feels too complex.  How about modelling on the Rust template?  https://github.com/rust-lang/rfcs/blob/master/0000-template.md

0000-template.md
```diff
@@ -0,0 +1,101 @@
...
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how Enarx contributors and/or users should *think* about the
+feature, and how it should impact the way they would use Enarx.
+- If applicable, provide sample error messages, deprecation warnings, or
+migration guidance.
+
+Some RFCs may be more aimed at contributors (e.g. for
+consensus internals); others may be more aimed at users.
```
connorkuehl: Might be a typo here, should we add the word "on" in between "consensus" and "internals" -> "consensus on internals"

0000-template.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Future Possibilities
+
+Think about what the natural extension and evolution of your proposal would be and how it would affect the project as a whole. Try to use this section as a tool to more fully consider all possible interactions with the project in your proposal. Also consider how the this all fits into the roadmap for the project and of the relevant sub-team.
+
+This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related.
+
+If you have tried and cannot think of any future possibilities, you may simply state that you cannot think of anything.
+
+Note that having something written down in the future-possibilities section is not a reason to accept the current or a future RFC; such notes should be in the section on motivation or rationale in this or subsequent RFCs. The section merely provides additional information.
```
connorkuehl: Minor nit: these line lengths are much longer than the rest of the lines in the document whereas the others break at a consistent column. This won't have any impact on how it's displayed after `markdown` passes over it, though.

contributing.md
```diff
@@ -0,0 +1,106 @@
+# Enarx RFCs
+
+## Contributing
+
+### Do you need an RFC?
+
+Use an RFC to advocate substantial changes to the Aries ecosystem, where
+those changes need to be understood by developers who *use* Aries. Minor
```
connorkuehl: ~~Aries~~ Enarx here as well

contributing.md
```diff
@@ -0,0 +1,106 @@
...
+changes are not RFC-worthy, and changes that are internal in nature,
+invisible to those consuming Aries, should be documented elsewhere.
```
connorkuehl: ~~Aries~~ Enarx

contributing.md
```diff
@@ -0,0 +1,106 @@
...
+Before writing an RFC, consider exploring the idea on the
+[chat](https://gitter.im/enarx/community), on a daily meeting call
+(see the [wiki page on contributing](
+https://github.com/enarx/enarx/wiki/How-to-contribute),
+Encouraging feedback from maintainers is a good sign that you're on
+the right track.
+
+### How to propose an RFC
+
+  - Fork [the RFC repo](https://github.com/enarx/rfc).
```
connorkuehl: ~~https://github.com/enarx/rfc~~ -> https://github.com/enarx/RFCs

(the former is 404 Not Found).

contributing.md
```diff
@@ -0,0 +1,106 @@
...
+  - Pick a descriptive folder name for your RFC. Don't pick a number yet.
+  - Decide which parent folder is appropriate for your RFC.
+    If it is about a specific protocol or decorator or feature, its parent
+    should be /features; if it is about a concept that will be used in many
```
connorkuehl: I think adding backticks around '/features' so it appears as `/features` helps it stand out as a technical component (file path, code snippet, etc) of the repo.

contributing.md
```diff
@@ -0,0 +1,106 @@
...
+    different features, its parent should be /concepts.
```
connorkuehl: Same comment here with '/concepts' -> `/concepts`

connorkuehl (changes requested):
- [x] Replace stray and forlorn instances of "Aries" with "Enarx"
- [x] Change link to point to the updated repo link (https://github.com/enarx/RFCs)

contributing.md
axelsimon: Indeed! Thanks.

0000-template.md
axelsimon: That's from the original, but it doesn't sound great. Using your wording, thanks.

0000-template.md
axelsimon: Yes, this is the part where i got really tired and knowing that it wouldn't make a difference in the MD rendering, pushed it so that everyone could start looking over it :) It's fixed now.

0000-template.md
```diff
@@ -0,0 +1,101 @@
+# Title (Ex. 0000: RFC Topic)
```
MikeCamel: I note that the Rust RFCs have moved well into the 5 digit range, so I suggest we plan ahead and start with the template as #00000.

We can only hope...

contributing.md
axelsimon: @npmccallum requested we rename the repo to `rfc` while we were discussing this PR on Gitter. This URL was to reflect that immediately, with plans to change the repo's URL after the PR went through :) I'm switching it to `enarx/rfcs` to align with [rust-lang/rfcs](https://github.com/rust-lang/rfcs/) and keep it all in lowercase.

contributing.md
axelsimon: Very good suggestion, thanks.

contributing.md
axelsimon: Agreed.

0000-template.md
axelsimon: :+1:

contributing.md
axelsimon: Yes, thanks, @lkatalin pointed out the same thing. I was more tired than i realised when i pushed this :)

contributing.md
axelsimon: `contributing.md` is supposed to be the file that helps you fill in the template file (0000-template.md). Rust doesn't use it though, only Aries, but it does feel too complex, i agree.  

I've tried to simplify it and make it a numbered list that is easy to follow and will actually help people fill the template.

README.md
```diff
@@ -1,2 +1,92 @@
...
+Many changes, including bug fixes and documentation improvements can be
+implemented and reviewed via the
+[normal GitHub pull request workflow](https://github.com/enarx/enarx/wiki/How-to-contribute).
+
+Some changes though are "substantial", and we ask that these be put through a
+bit of a design process and produce a consensus among the community and
+the [sub-team]s.
+
+If you are here to learn about Enarx, as the code is still young, we
+recommend you look at the [code][https://github.com/enarx/enarx/) and
```
mbestavros:
```suggestion
recommend you look at the [code](https://github.com/enarx/enarx/) and
```

Typo. The link would break.

README.md
```diff
@@ -1,2 +1,92 @@
-# RFCs
-Holds RFCS (“Request for Comments”), proposals to improve Enarx
+# RFCs for Enarx
+
+This repo holds Request for Comments (RFCs), enhancement proposals to improve
+the Enarx project.
+
+Many changes, including bug fixes and documentation improvements can be
```
mbestavros:
```suggestion
Many changes, including bug fixes and documentation improvements, can be
```

README.md
```diff
@@ -1,2 +1,92 @@
...
+implemented and reviewed via the
+[normal GitHub pull request workflow](https://github.com/enarx/enarx/wiki/How-to-contribute).
+
+Some changes though are "substantial", and we ask that these be put through a
```
mbestavros:
```suggestion
Some changes, though, are "substantial", and we ask that these be put through a
```

README.md
```diff
@@ -1,2 +1,92 @@
...
+
+## RFC Lifecycle
+
+RFCs go through a standard lifecycle.
+
+![lifecycle](lifecycle.png)
+
+#### PROPOSED
+To __propose__ an RFC, [use these instructions to raise a PR](
+contributing.md#how-to-propose-an-RFC) (Pull Request) against the repo.
```
mbestavros: Would the line break here break the link?

00000-template.md
```diff
@@ -0,0 +1,110 @@
...
+RFC process before this gets merged?
+- What parts of the design do you expect to resolve through the
+implementation of this feature before stabilization?
+- What related issues do you consider out of scope for this 
+proposal that could be addressed in the future independently of the
+solution that comes out of this doc?
+
+## Implementations
+
+If your proposal is related to code and you have a early implementation,
```
mbestavros:
```suggestion
If your proposal is related to code and you have an early implementation,
```

mbestavros (commented):
Mostly grammatical nitpicks from me. Otherwise, this process makes a lot of sense. The structure the template provides will be super helpful for fleshing out a detailed proposal -- I find that's often something that holds me up.

If anything, I think I'll echo the sentiments I've heard saying it's important to ensure the PR process isn't too cumbersome. That should not be something that takes a whole lot of time.

Perhaps something we could consider for the RFC number is just using the PR number assigned by Github (ex. this one would be `00001`, since it's PR #1 on the repo.)

I will defer approval to others, but I think this is a good start.

README.md
```diff
@@ -1,2 +1,92 @@
...
+This repo holds Request for Comments (RFCs), enhancement proposals to improve
+the Enarx project.
+
+Many changes, including bug fixes and documentation improvements can be
+implemented and reviewed via the
+[normal GitHub pull request workflow](https://github.com/enarx/enarx/wiki/How-to-contribute).
+
+Some changes though are "substantial", and we ask that these be put through a
+bit of a design process and produce a consensus among the community and
+the [sub-team]s.
```
connorkuehl: Is there supposed to be a link here for this sub-team text?

README.md
connorkuehl: ~~use these instructions to raise a PR (Pull Request)~~ -> use these instructions to raise a Pull Request (PR)

connorkuehl: It doesn't appear to break in my markdown preview in VS Code.

README.md
```diff
@@ -1,2 +1,92 @@
...
+#### DEMONSTRATED
+__Demonstrated__ RFCs have one or more implementations available, listed in
+the "Implementations" section of the RFC document. As with the PROPOSED status,
+demonstrated RFCs haven't been endorsed by the community, but the ideas put
+forth have been more thoroughly explored through the implementation(s).
+The demonstrated status is an __optional__ step in the lifecycle.
+
+#### ACCEPTED
+To get an RFC __accepted__, [build consensus](contributing.md#how-to-get-an-RFC-accepted)
+for your RFC on [chat](https://gitter.im/enarx/community) and in community
```
connorkuehl: ~~chat~~ -> gitter

README.md
```diff
@@ -1,2 +1,92 @@
...
+meetings. If your RFC is a feature that's code-related, it MUST have
+reasonable tests. An accepted RFC is incubating on a standards track;
+the community has decided to polish it and is exploring or pursuing implementation.
+
+#### ADOPTED
+To get an RFC __adopted__, [socialize and implement](contributing.md#how-to-get-an-rfc-adopted).
```
connorkuehl: I think we should put the word "it" after "implement"

README.md
```diff
@@ -1,2 +1,92 @@
...
+
+### Changing an RFC Status
+
+See notes about this in [Contributing](contributing.md#changing-an-rfc-status).
+
+## About
+
+#### License
+
+This repository is licensed under an [Apache 2 License](LICENSE).
```
connorkuehl: I'm having flashbacks to online math assignment submission portals, but I wonder if it should be Apache 2.0 instead of Apache 2... I'm not as familiar with Apache so I don't know if there already is or will be things like Apache 2.1, etc or if that's even important...

contributing.md
```diff
@@ -0,0 +1,113 @@
...
+- Update the agent test suite to reflect the ideas.
+
+When you believe an RFC is a _de facto_ standard, [raise a PR](#changing-an-rfc-status) that changes the
+status to [ADOPTED](README.md#status--adopted).  If the community is friendly
+to the idea, the doc will enter a two-week "Final Comment Period" (FCP), after
+which there will be a vote on disposition.
+
+### Intellectual Rights
+
+This repository is licensed under an [Apache 2 License](LICENSE).
```
connorkuehl: Same 2 vs 2.0 comment as above.

README.md
mbestavros: I think I agree. Apache 2.0...

README.md
axelsimon: Thanks.

README.md
axelsimon: Thanks!

README.md
axelsimon: Thanks!

00000-template.md
axelsimon: Thanks!

README.md
axelsimon: Good catch, thanks. Rust and Aries have sub-teams, we sort of do too but in a less formal way. I'll remove the half link for now, we can always re-add later.

README.md
axelsimon: Thanks. I'll avoid the line break and opt for a longer line in the case of that link.
I'd rather keep chat than Gitter, as we are using Gitter for now and it's working fine, but could change in the future and chat is generic while being specific enough for our purposes here.

README.md
axelsimon: Makes sense, adding this, thanks.

README.md
axelsimon: Agreed, thanks. Nice catch.

axelsimon:
I've incorporated all suggested changes and pushed a new version of the PR.

Many thanks to all of you for taking the time to read through these documents as carefully as you did and making suggestions and corrections, it's really nice to see everyone getting behind this RFC process :)

axelsimon:
@connorkuehl, @mbestavros you were the last to have requested changes or made comments in that general direction. Does everything look good to you? I'd like to merge this if so :)

axelsimon:
Pinging @npmccallum too as he expressed a strong interest in checking this PR out.

connorkuehl (approved):
I like this very much. Thank you, @axelsimon!

npmccallum (changes requested):
This is all very good. Thank you to everyone who contributed. In fact, I'm pleased to see so many directly involved.

I particularly liked @mbestavros's idea of using the PR# for the RFC number. This guarantees both uniqueness and numeric increase. It also means that we can write a test to validate the correct RFC number to prevent human error. Therefore, this is the only change I'd like to see adopted at this point.

axelsimon:
Funny you should mention that as i was at a matrix meetup yesterday and as i realised that their [MSC](https://github.com/matrix-org/matrix-doc/pulls) (matrix spec changes) process uses this same system, and thought we should do the same.
So yes, very much agreed. Will make the needed changes and then push this.

PR #2, opened 2020-02-10
Wasm rfc 0001

MikeCamel:
Initial draft of workload loading RFC.
Created two new issues based on the work: #227, #228.

Added line breaks: what are the kids using as default line length these days? 

Addresses issue #140.

### DIFF
--- a/rfc#0001
+++ b/rfc#0001
@@ -62,8 +62,8 @@ Some TEE implementations assume trust in the host system.  This is explicitly ex
 1. What should the well-known port number be?  Should we use HTTPS (443) or another?
 
 Out of scope:
-1. Keep preparation steps.
-1. Workload package format.
+1. Keep preparation steps: requires RFC.
+1. Workload package format: requires RFC.
 
 ## Implementations
 
--- a/rfc#0001
+++ b/rfc#0001
@@ -22,7 +22,7 @@ description of the loading sequence is required.
 
 A Keep is the fundamental runtime component of Enarx: a TEE instance
 + Enarx runtime pieces, including WebAssembly and WASI layers.  In
-order for it to execute a workload, the Wasm (WebAssembly) package
+order for the Keep to execute a workload, the Wasm (WebAssembly) package
 that comprises it (binary/binaries + data elements) need to be loaded
 into it.  This is sent to it by the Enarx client agent (ECA),
 encrypted under a session key established at Keep provisioning time,
--- a/rfc#0001
+++ b/rfc#0001
@@ -128,7 +128,7 @@ handle key management and decryption/loading of the workload.
 above 1024 may make life a little easier (non-root binding).
 
 Out of scope:
-1. Keep preparation steps: requires RFC - issue #228.
+1. Keep preparation steps: requires RFC - issue #16.
 1. Workload package format: requires RFC - issue #227.
 
 ## Implementations
--- a/rfc#0001
+++ b/rfc#0001
@@ -31,7 +31,7 @@ The Keep has an agent running which is listening on a
 well-known port, or on a port number previously established during
 the Keep preparation phase.  The ECA connects to this port, and
 establishes a connection to the Keep.  The Keep takes the workload,
-decrypts it, and runs it.
+decrypts it with the WPEK, and runs it.
 
 For this design to work, a route between the ECA and the Keep control
 plane needs to be available.  Note that this does not necessarily
--- a/rfc#0001
+++ b/rfc#0001
@@ -24,9 +24,11 @@ A Keep is the fundamental runtime component of Enarx: a TEE instance
 + Enarx runtime pieces, including WebAssembly and WASI layers.  In
 order for the Keep to execute a workload, the Wasm (WebAssembly) package
 that comprises the workload (binary/binaries + data elements) need to be loaded
-into the Keep.  This workload is sent to the Keep by the Enarx client agent (ECA),
-encrypted under a session key (the workload provisioning encryption key or WPEK) that is established at Keep provisioning time
-and communicated to the ECA.
+into the Keep.  This workload is sent to the Keep by the Enarx client agent
+(ECA), encrypted under a session key (the workload provisioning encryption
+key or WPEK) that is established at Keep provisioning time and
+communicated to the ECA.
+
 The Keep has an agent running which is listening on a
 well-known port, or on a port number previously established during
 the Keep preparation phase.  The ECA connects to this port, and
@@ -108,8 +110,10 @@ and runs. This maintains the host agent's status as a proxy between
 other components, and doesn't make further assumptions about routing
 which is made by 2. 
 Rejected:
-- the host agent should not need logic to provide functionality.
-- maintaining multiple versions is complex.
+- the host agent should act as an untrusted proxy wherever possible,
+and not include process logic.
+- maintaining multiple versions of host agents is complex from
+a deployment point of view.
 
 ## Prior art
 
--- a/rfc#0001
+++ b/rfc#0001
@@ -16,7 +16,7 @@ packages into a prepared Keep.
 
 Once a Keep is prepared with microkernel, WASM runtime, etc., the
 workload (WASM binary + data) needs to be loaded into it.  A
-description is required.
+description of the loading sequence is required.
 
 ## Tutorial
 
--- a/rfc#0001
+++ b/rfc#0001
@@ -9,20 +9,20 @@
 
 ## Summary
 
-This RFC describes the architecture adopted by Enarx for loading WASM
+This RFC describes the architecture adopted by Enarx for loading Wasm
 packages into a prepared Keep. 
 
 ## Motivation
 
-Once a Keep is prepared with microkernel, WASM runtime, etc., the
-workload (WASM binary + data) needs to be loaded into it.  A
-description is required.
+Once a Keep is prepared with microkernel, Wasm runtime, etc., the
+tenant workload (Wasm binary + data) needs to be loaded into it.  A
+description of the loading sequence is required.
 
 ## Tutorial
 
 A Keep is the fundamental runtime component of Enarx: a TEE instance
 + Enarx runtime pieces, including WebAssembly and WASI layers.  In
-order for it to execute a workload, the WASM (WebAssembly) package
+order for it to execute a workload, the Wasm (WebAssembly) package
 that comprises it (binary/binaries + data elements) need to be loaded
 into it.  This is sent to it by the Enarx client agent (ECA),
 encrypted under a session key established at Keep provisioning time,
@@ -56,7 +56,7 @@ transfer to another Keep in the same trust domain.  This caching, if
 under the origincal WPEK, is at the expense of Perfect Forward
 Secrecy.  If a Keep is to transfer the workload to another Keep in
 the same domain, it MAY encrypt it under a new WPEK.  If the
-original is used, it will need to established in the new Keep either
+original is used, it will need to be established in the new Keep either
 at provisioning time (e.g. by the ECA) or at a later date through
 key negotiation.  Such functionality is considered risky, as the key
 management required is difficult to get right, and provides new
@@ -84,7 +84,7 @@ enable caching.
 
 ## Reference
 
-Empty.
+This RFC arose from this issue: [Workload (WASM binary) loading - connection termination](https://github.com/enarx/enarx/issues/140)
 
 ## Drawbacks
 
--- a/rfc#0001
+++ b/rfc#0001
@@ -84,7 +84,8 @@ enable caching.
 
 ## Reference
 
-This RFC arose from this issue: [Workload (WASM binary) loading - connection termination](https://github.com/enarx/enarx/issues/140)
+This RFC arose from this issue: [Workload (WASM binary) loading
+- connection termination](https://github.com/enarx/enarx/issues/140)
 
 ## Drawbacks
 
--- a/rfc#0001
+++ b/rfc#0001
@@ -50,7 +50,7 @@ server.  However, as the workload provisioning encryption key (WPEK)
 has already been communicated to the Keep, if the workload can be
 decrypted using this key, this assures the Keep that it is
 communicating with the ECA, or at least an entity in the same trust
-domain, so identity has been confirmed.
+domain, so tenant identity has been confirmed.
 
 There are possible use cases where a Keep might wish to cache a copy
 of the workload, either for reloading ("recycling" a Keep) or to
--- a/rfc#0001
+++ b/rfc#0001
@@ -23,7 +23,7 @@ description of the loading sequence is required.
 A Keep is the fundamental runtime component of Enarx: a TEE instance
 + Enarx runtime pieces, including WebAssembly and WASI layers.  In
 order for the Keep to execute a workload, the Wasm (WebAssembly) package
-that comprises it (binary/binaries + data elements) need to be loaded
+that comprises the workload (binary/binaries + data elements) need to be loaded
 into the Keep.  This workload is sent to the Keep by the Enarx client agent (ECA),
 encrypted under a session key (the workload provisioning encryption key or WPEK) that is established at Keep provisioning time
 and communicated to the ECA.
--- a/rfc#0001
+++ b/rfc#0001
@@ -24,7 +24,7 @@ A Keep is the fundamental runtime component of Enarx: a TEE instance
 + Enarx runtime pieces, including WebAssembly and WASI layers.  In
 order for the Keep to execute a workload, the Wasm (WebAssembly) package
 that comprises it (binary/binaries + data elements) need to be loaded
-into it.  This is sent to it by the Enarx client agent (ECA),
+into the Keep.  This workload is sent to the Keep by the Enarx client agent (ECA),
 encrypted under a session key established at Keep provisioning time,
 and communicated to the ECA - the workload provisioning encryption
 key (WPEK).  The Keep has an agent running which is listening on a
--- a/rfc#0001
+++ b/rfc#0001
@@ -9,29 +9,78 @@
 
 ## Summary
 
-This RFC describes the architecture adopted by Enarx for loading WASM packages into a prepared Keep. 
+This RFC describes the architecture adopted by Enarx for loading WASM
+packages into a prepared Keep. 
 
 ## Motivation
 
-Once a Keep is prepared with microkernel, WASM runtime, etc., the workload (WASM binary + data) needs to be loaded into it.  A description is required.
+Once a Keep is prepared with microkernel, WASM runtime, etc., the
+workload (WASM binary + data) needs to be loaded into it.  A
+description is required.
 
 ## Tutorial
 
-A Keep is the fundamental runtime component of Enarx: a TEE instance + Enarx runtime pieces, including WebAssembly and WASI layers.  In order for it to execute a workload, the WASM (WebAssembly) package that comprises it (binary/binaries + data elements) need to be loaded into it.  This is sent to it by the Enarx client agent (ECA), encrypted under a session key established at Keep provisioning time, and communicated to the ECA - the workload provisioning encryption key (WPEK).  The Keep has an agent running which is listening on a well-known port, or on a port number previously established during the Keep preparation phase.  The ECA connects to this port, and establishes a connection to the Keep.  The Keep takes the workload, decrypts it, and runs it.
+A Keep is the fundamental runtime component of Enarx: a TEE instance
++ Enarx runtime pieces, including WebAssembly and WASI layers.  In
+order for it to execute a workload, the WASM (WebAssembly) package
+that comprises it (binary/binaries + data elements) need to be loaded
+into it.  This is sent to it by the Enarx client agent (ECA),
+encrypted under a session key established at Keep provisioning time,
+and communicated to the ECA - the workload provisioning encryption
+key (WPEK).  The Keep has an agent running which is listening on a
+well-known port, or on a port number previously established during
+the Keep preparation phase.  The ECA connects to this port, and
+establishes a connection to the Keep.  The Keep takes the workload,
+decrypts it, and runs it.
 
-For this design to work, a route between the ECA and the Keep control plane needs to be available.  Note that this does not necessarily need to be direct: since an encrypted connection is not explicitly required (see below), proxying is possible.  This could, if required, be via an (untrusted) service on the host.
+For this design to work, a route between the ECA and the Keep control
+plane needs to be available.  Note that this does not necessarily
+need to be direct: since an encrypted connection is not explicitly
+required (see below), proxying is possible.  This could, if required,
+be via an (untrusted) service on the host.
 
-One of the Enarx project's design principles is that all network and storage connections in and out of a Keep are encrypted.  In the case of network connections, this usually means that the session is encrypted using TLS (or dTLS), but since we may not have established any certificates in the Keep at this point, it cannot act as the server.  However, as the workload provisioning encryption key (WPEK) has already been communicated to the Keep, if the workload can be decrypted using this key, this assures the Keep that it is communicating with the ECA, or at least an entity in the same trust domain, so identity has been confirmed.
+One of the Enarx project's design principles is that all network
+and storage connections in and out of a Keep are encrypted.  In the
+case of network connections, this usually means that the session is
+encrypted using TLS (or dTLS), but since we may not have established
+any certificates in the Keep at this point, it cannot act as the
+server.  However, as the workload provisioning encryption key (WPEK)
+has already been communicated to the Keep, if the workload can be
+decrypted using this key, this assures the Keep that it is
+communicating with the ECA, or at least an entity in the same trust
+domain, so identity has been confirmed.
 
-There are possible use cases where a Keep might wish to cache a copy of the workload, either for reloading ("recycling" a Keep) or to transfer to another Keep in the same trust domain.  This caching, if under the origincal WPEK, is at the expense of Perfect Forward Secrecy.  If a Keep is to transfer the workload to another Keep in the same domain, it MAY encrypt it under a new WPEK.  If the original is used, it will need to established in the new Keep either at provisioning time (e.g. by the ECA) or at a later date through key negotiation.  Such functionality is considered risky, as the key management required is difficult to get right, and provides new options for compromise or design/implementation errors.  Any design of such functionality requires a new RFC (see "Future possibilities" below).
+There are possible use cases where a Keep might wish to cache a copy
+of the workload, either for reloading ("recycling" a Keep) or to
+transfer to another Keep in the same trust domain.  This caching, if
+under the origincal WPEK, is at the expense of Perfect Forward
+Secrecy.  If a Keep is to transfer the workload to another Keep in
+the same domain, it MAY encrypt it under a new WPEK.  If the
+original is used, it will need to established in the new Keep either
+at provisioning time (e.g. by the ECA) or at a later date through
+key negotiation.  Such functionality is considered risky, as the key
+management required is difficult to get right, and provides new
+options for compromise or design/implementation errors.  Any design
+of such functionality requires a new RFC (see "Future possibilities"
+below).
 
-The Keep MUST NOT accept unencrypted workloads, as there is no way to verify their provenance.
-The Keep SHOULD listen on a well-known port (see "Unresolved questions").
-If a port has been pre-esablished, the Keep MUST listen on this port, and MAY also listen on the well-known port.
-The Keep MAY cache the workload for re-use, or transfer to another Keep.  If a Keep allows workloads to be used by another Keep, the workload MAY be encrypted under a new WPEK, established between the original Keep and new Keep.
-Keeps MAY be implemented to enable the capability to performing workload caching.
-Enabling of the capability for a Keep to perform caching MUST be at the discretion of the ECA or entities within the ECA's trust domain.
-Enabling of the capability for a Keep to perform caching MUST NOT be at the discretion of the host.  The host MUST NOT be able to enable caching.
+The Keep MUST NOT accept unencrypted workloads, as there is no way
+to verify their provenance.
+The Keep SHOULD listen on a well-known port (see "Unresolved
+questions").
+If a port has been pre-esablished, the Keep MUST listen on this port,
+and MAY also listen on the well-known port.
+The Keep MAY cache the workload for re-use, or transfer to another
+Keep.  If a Keep allows workloads to be used by another Keep, the
+workload MAY be encrypted under a new WPEK, established between the
+original Keep and new Keep.
+Keeps MAY be implemented to enable the capability to performing
+workload caching.
+Enabling of the capability for a Keep to perform caching MUST be at
+the discretion of the ECA or entities within the ECA's trust domain.
+Enabling of the capability for a Keep to perform caching MUST NOT
+be at the discretion of the host.  The host MUST NOT be able to
+enable caching.
 
 ## Reference
 
@@ -45,21 +94,33 @@ Why should we *not* do this?
 
 Alternatives considered:
 
-Enarx client agent encrypts workload, sends to CPU/firmware via Enarx host agent. CPU/firmware decrypts workload, loads into Keep. This has the advantage that we use an existing network connection, and there's no need to establish a connection nor a session key between the Keep and the client agent.
+Enarx client agent encrypts workload, sends to CPU/firmware via Enarx
+host agent. CPU/firmware decrypts workload, loads into Keep. This has
+the advantage that we use an existing network connection, and there's
+no need to establish a connection nor a session key between the Keep
+and the client agent.
 Rejected: this is not supported by all architectures.
 
-Enarx client agent encrypts workload, sends to host agent. Host agent sends to Keep via a listening agent in the Keep. Keep decrypts, loads and runs. This maintains the host agent's status as a proxy between other components, and doesn't make further assumptions about routing which is made by 2. 
+Enarx client agent encrypts workload, sends to host agent. Host agent
+sends to Keep via a listening agent in the Keep. Keep decrypts, loads
+and runs. This maintains the host agent's status as a proxy between
+other components, and doesn't make further assumptions about routing
+which is made by 2. 
 Rejected:
 - the host agent should not need logic to provide functionality.
 - maintaining multiple versions is complex.
 
 ## Prior art
 
-Some TEE implementations assume trust in the host system.  This is explicitly excluded as one of the Enarx project's design principles.  In such implementations, a component trusted by the host system may handle key management and decryption/loading of the workload.
+Some TEE implementations assume trust in the host system.  This is
+explicitly excluded as one of the Enarx project's design principles.
+In such implementations, a component trusted by the host system may
+handle key management and decryption/loading of the workload.
 
 ## Unresolved questions
 
-1. What should the well-known port number be?  Should we use HTTPS (443) or another?
+1. What should the well-known port number be?  Should we use HTTPS
+(443) or another?
 
 Out of scope:
 1. Keep preparation steps: requires RFC - issue #228.
@@ -73,4 +134,7 @@ No implementation yet available (2020-02-10).
 
 Caching designs should be considered under a new RFC.
 
-There are opportunities for "pre-warmed" Keeps, which are provisioned ahead of use time, and whose WPEK is not necessarily derived from the initial session key established between the ECA and the waiting Keep.  Designs for such should be under a new RFC.
+There are opportunities for "pre-warmed" Keeps, which are provisioned
+ahead of use time, and whose WPEK is not necessarily derived from the
+initial session key established between the ECA and the waiting Keep.
+Designs for such should be under a new RFC.
--- a/rfc#0001
+++ b/rfc#0001
@@ -23,7 +23,7 @@ For this design to work, a route between the ECA and the Keep control plane need
 
 One of the Enarx project's design principles is that all network and storage connections in and out of a Keep are encrypted.  In the case of network connections, this usually means that the session is encrypted using TLS (or dTLS), but since we may not have established any certificates in the Keep at this point, it cannot act as the server.  However, as the workload provisioning encryption key (WPEK) has already been communicated to the Keep, if the workload can be decrypted using this key, this assures the Keep that it is communicating with the ECA, or at least an entity in the same trust domain, so identity has been confirmed.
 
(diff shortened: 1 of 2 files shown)


rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+## Summary
+
+This RFC describes the architecture adopted by Enarx for loading WASM
+packages into a prepared Keep. 
+
+## Motivation
+
+Once a Keep is prepared with microkernel, WASM runtime, etc., the
+workload (WASM binary + data) needs to be loaded into it.  A
+description is required.
```
steveej:
```suggestion
description of the loading sequence is required.
```

rfc#0001
```diff
@@ -0,0 +1,140 @@
+# WASM binary loading into an Enarx Keep
```
steveej: I recently learned that *Wasm* is the correct abbreviation of WebAssembly:
> A contraction of “WebAssembly”, not an acronym, hence not using all-caps. 

--  [WebAssembly draft spec](https://webassembly.github.io/spec/core/bikeshed/index.html#id1)

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+
+## Summary
+
+This RFC describes the architecture adopted by Enarx for loading WASM
+packages into a prepared Keep. 
+
+## Motivation
+
+Once a Keep is prepared with microkernel, WASM runtime, etc., the
+workload (WASM binary + data) needs to be loaded into it.  A
```
steveej: might we want to say "tenant workload" here to be very specific?

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+order for it to execute a workload, the WASM (WebAssembly) package
+that comprises it (binary/binaries + data elements) need to be loaded
+into it.  This is sent to it by the Enarx client agent (ECA),
+encrypted under a session key established at Keep provisioning time,
+and communicated to the ECA - the workload provisioning encryption
+key (WPEK).  The Keep has an agent running which is listening on a
+well-known port, or on a port number previously established during
+the Keep preparation phase.  The ECA connects to this port, and
+establishes a connection to the Keep.  The Keep takes the workload,
+decrypts it, and runs it.
```
steveej: May there be use-cases where it decrypts it but doesn't run it immediately? E.g., might the tenant want to measure the Keep after loading the workload but before running it?

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+communicating with the ECA, or at least an entity in the same trust
+domain, so identity has been confirmed.
+
+There are possible use cases where a Keep might wish to cache a copy
+of the workload, either for reloading ("recycling" a Keep) or to
+transfer to another Keep in the same trust domain.  This caching, if
+under the origincal WPEK, is at the expense of Perfect Forward
+Secrecy.  If a Keep is to transfer the workload to another Keep in
+the same domain, it MAY encrypt it under a new WPEK.  If the
+original is used, it will need to established in the new Keep either
```
steveej:
```suggestion
original is used, it will need to be established in the new Keep either
```

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+to verify their provenance.
+The Keep SHOULD listen on a well-known port (see "Unresolved
+questions").
+If a port has been pre-esablished, the Keep MUST listen on this port,
+and MAY also listen on the well-known port.
+The Keep MAY cache the workload for re-use, or transfer to another
+Keep.  If a Keep allows workloads to be used by another Keep, the
+workload MAY be encrypted under a new WPEK, established between the
+original Keep and new Keep.
+Keeps MAY be implemented to enable the capability to performing
```
steveej: I'm never sure what people mean by *implement*. Does this merely mean configured in this context?

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+workload caching.
+Enabling of the capability for a Keep to perform caching MUST be at
+the discretion of the ECA or entities within the ECA's trust domain.
+Enabling of the capability for a Keep to perform caching MUST NOT
+be at the discretion of the host.  The host MUST NOT be able to
```
steveej: Shall we discuss the design to protect this configuration from being changed by the host within this RFC?

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+enable caching.
+
+## Reference
+
+Empty.
```
steveej: You could link the GitHub issue which preceded this RFC.

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+at provisioning time (e.g. by the ECA) or at a later date through
+key negotiation.  Such functionality is considered risky, as the key
+management required is difficult to get right, and provides new
+options for compromise or design/implementation errors.  Any design
+of such functionality requires a new RFC (see "Future possibilities"
+below).
+
+The Keep MUST NOT accept unencrypted workloads, as there is no way
+to verify their provenance.
+The Keep SHOULD listen on a well-known port (see "Unresolved
```
steveej: This involves the host to bind this well-known port and forward all incoming packets to the Keep in question. I'm not sure if I suggest the wording here to include this fact.

rfc#0001
```diff
@@ -0,0 +1,140 @@
...
+## Prior art
+
+Some TEE implementations assume trust in the host system.  This is
+explicitly excluded as one of the Enarx project's design principles.
+In such implementations, a component trusted by the host system may
+handle key management and decryption/loading of the workload.
+
+## Unresolved questions
+
+1. What should the well-known port number be?  Should we use HTTPS
```
steveej: I suggest to use a port >1024 so that we can bind the ports on the host as a non-root user. Which one specifically doesn't seem important, and any unassigned by IANA will do.

steveej (commented):
Thanks for taking the time to write this up!

rfc#0001
MikeCamel: Changing throughout: thanks.

rfc#0001
MikeCamel: Added.

rfc#0001
MikeCamel: I don't this is excluded by the text as it stands.  However, I don't think your particular use case is likely: the tenant should not send the workload to the Keep until it is confident that it has been successfully attested.

rfc#0001
MikeCamel: That depends rather on the networking set-up that we're assuming.  I'm guessing by this point that the Keep already has an IP address and route to it, so binding isn't an issue hopefully.

rfc#0001
MikeCamel: No, it means designed and coded.  This should become an RFC document - a design template, if you will - which allows different implementations to exist which are all compliant with this RFC, but which exhibit different behaviour.  Some might be newer versions, some might exist only on particular architectures, and others might be clean-room implementations.  When we actually create a design doc, we'll specify exact behaviour for the implementation that design describes.

Think of this RFC as a standards doc like an IETF RFC for SMTP, DNS, etc..

Note: I'd like to capture the above in the template, @axelsimon.

rfc#0001
MikeCamel: I don't believe so, for the reasons above that this is an RFC, not a design.

rfc#0001
MikeCamel: Done.

rfc#0001
MikeCamel: Except that HTTPS is on 443.  Certainly requires more discussion.

rfc#0001
steveej: Thanks for elaborating. It shows me that I have some getting accustomed to do for the RFC language :-)

rfc#0001
steveej: I'm personally interested to have the Keep state verified after the workload is loaded. But I think it doesn't have to be part of this RFC. How about removing anything from this RFC after the workload has been received and decrypted by the Keep?

rfc#0001
steveej: Instead of speaking about ports which sounds like implementation language to me, could we use a more general term, e.g. communication endpoint?

rfc#0001
steveej: I don't know of any plans to use HTTPS, specifically on its default port. Maybe it can be left out of the RFC, and _well-known communication endpoint_ might suffice to specify the behavior. See https://github.com/enarx/rfcs/pull/2#discussion_r377204173

rfc#0001
MikeCamel: I think we're specifically talking network ports for this design.  We can certainly discuss, however.  And that doesn't mean that our initial implementation will look like that.

rfc#0001
MikeCamel: This is here as a discussion point, and we may well remove it.  Given text elsewhere about _not_ encrypting at the connection level, HTTPS may not make sense.

rfc#0001
MikeCamel: Do you mean the tenant workload?  That's what this RFC is referring to.  If so, that's difficult: though we might be able to measure the Keep+workload, we are unlikely to have a reference measurement against which to check it.

rfc#0001
steveej: @axelsimon I'll leave the resolving of this comment to you to acknowledge Mike's mention.

MikeCamel: Discussed with @axelsimon .

rfc#0001
MikeCamel: Changed this from HTTPS, but left it with "port number" see notes above.

rfc#0001
MikeCamel: Would like @npmccallum to weigh in here.  There's a balance between over-vagueness and enough for us to move forward with a good implementation.  We've always assumed network connections, so unless there's a good reason to move away from this, that's my preference.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+
+Once a Keep is prepared with microkernel, Wasm runtime, etc., the
+tenant workload (Wasm binary + data) needs to be loaded into it.  A
+description of the loading sequence is required.
+
+## Tutorial
+
+A Keep is the fundamental runtime component of Enarx: a TEE instance
++ Enarx runtime pieces, including WebAssembly and WASI layers.  In
+order for it to execute a workload, the Wasm (WebAssembly) package
```
lkatalin:
```suggestion
order for the Keep to execute a workload, the Wasm (WebAssembly) package
```
Best to be very specific and avoid pronouns wherever possible when there are so many technical pieces being described, IMO.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+that comprises it (binary/binaries + data elements) need to be loaded
```
lkatalin:
```suggestion
that comprises the workload (binary/binaries + data elements) need to be loaded
```
Is this right?

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+into it.  This is sent to it by the Enarx client agent (ECA),
```
lkatalin:
```suggestion
into the Keep.  This workload is sent to the Keep by the Enarx client agent (ECA),
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+encrypted under a session key established at Keep provisioning time,
```
lkatalin:
```suggestion
encrypted under a session key (the workload provisioning encryption key or WPEK) that is established at Keep provisioning time
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+and communicated to the ECA - the workload provisioning encryption
```
lkatalin:
```suggestion
and communicated to the ECA.
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+key (WPEK).  The Keep has an agent running which is listening on a
```
lkatalin:
```suggestion
The Keep has an agent running which is listening on a
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+well-known port, or on a port number previously established during
+the Keep preparation phase.  The ECA connects to this port, and
+establishes a connection to the Keep.  The Keep takes the workload,
+decrypts it, and runs it.
+
+For this design to work, a route between the ECA and the Keep control
+plane needs to be available.  Note that this does not necessarily
```
lkatalin: Is the Keep control plane necessarily within the Keep? Is an untrusted service on the host conceptually separate from the Keep control plane? (It makes sense that it should be since it's not trusted, but maybe a few words defining the extent of the control plane could be useful.)

rfc#0001
lkatalin: The WPEK is communicated to the ECA by whom? Established by what entities?

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+required (see below), proxying is possible.  This could, if required,
+be via an (untrusted) service on the host.
+
+One of the Enarx project's design principles is that all network
+and storage connections in and out of a Keep are encrypted.  In the
+case of network connections, this usually means that the session is
+encrypted using TLS (or dTLS), but since we may not have established
+any certificates in the Keep at this point, it cannot act as the
+server.  However, as the workload provisioning encryption key (WPEK)
+has already been communicated to the Keep, if the workload can be
```
lkatalin: > has already been communicated to the Keep

I think this needs to be explicitly stated in the WPEK section above. Currently that section states the ECA has access to the key, but other parties are only implied. The WPEK is actually said to be communicated to both the Keep and the ECA, but what entity is communicating this? I think it should be explicit that (except in the case of caching) it's a Diffie-Hellman so that it is clear that other parties do not know the WPEK and that there's not a general-purpose WPEK being used for all Keeps.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+decrypted using this key, this assures the Keep that it is
+communicating with the ECA, or at least an entity in the same trust
```
lkatalin: > this assures the Keep that it is communicating with the ECA,

Wouldn't this be the other way around? The ECA wants to know that it is communicating with the same attested Keep? Does the Keep care who it is communicating with as well?

Edit: From below paragraphs, it sounds like the Keep does care. I think I just missed this.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+domain, so identity has been confirmed.
+
+There are possible use cases where a Keep might wish to cache a copy
+of the workload, either for reloading ("recycling" a Keep) or to
+transfer to another Keep in the same trust domain.  This caching, if
+under the origincal WPEK, is at the expense of Perfect Forward
+Secrecy.  If a Keep is to transfer the workload to another Keep in
+the same domain, it MAY encrypt it under a new WPEK.  If the
+original is used, it will need to be established in the new Keep either
```
lkatalin:
```suggestion
original WPEK is used, that WPEK will need to be established in the new Keep either
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+server.  However, as the workload provisioning encryption key (WPEK)
+has already been communicated to the Keep, if the workload can be
+decrypted using this key, this assures the Keep that it is
+communicating with the ECA, or at least an entity in the same trust
+domain, so identity has been confirmed.
+
+There are possible use cases where a Keep might wish to cache a copy
+of the workload, either for reloading ("recycling" a Keep) or to
+transfer to another Keep in the same trust domain.  This caching, if
+under the origincal WPEK, is at the expense of Perfect Forward
```
lkatalin:
```suggestion
under the original WPEK, is at the expense of Perfect Forward
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+Secrecy.  If a Keep is to transfer the workload to another Keep in
+the same domain, it MAY encrypt it under a new WPEK.  If the
```
lkatalin:
```suggestion
the same domain, the original Keep MAY encrypt the workload under a new WPEK.  If the
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+original is used, it will need to be established in the new Keep either
+at provisioning time (e.g. by the ECA) or at a later date through
+key negotiation.  Such functionality is considered risky, as the key
+management required is difficult to get right, and provides new
+options for compromise or design/implementation errors.  Any design
+of such functionality requires a new RFC (see "Future possibilities"
+below).
+
+The Keep MUST NOT accept unencrypted workloads, as there is no way
+to verify their provenance.
```
lkatalin: I guess the Keep does care where its workloads come from... disregard the previous comment about this, probably.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+The Keep SHOULD listen on a well-known port (see "Unresolved
+questions").
+If a port has been pre-esablished, the Keep MUST listen on this port,
```
lkatalin:
```suggestion
If a port has been pre-established between the ECA and the Keep during the Keep preparation phase, the Keep MUST listen on this port,
```

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+Keep.  If a Keep allows workloads to be used by another Keep, the
+workload MAY be encrypted under a new WPEK, established between the
+original Keep and new Keep.
+Keeps MAY be implemented to enable the capability to performing
+workload caching.
+Enabling of the capability for a Keep to perform caching MUST be at
+the discretion of the ECA or entities within the ECA's trust domain.
+Enabling of the capability for a Keep to perform caching MUST NOT
+be at the discretion of the host.  The host MUST NOT be able to
+enable caching.
```
lkatalin: > The host MUST NOT be able to enable caching.

Strongly agree.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+and the client agent.
+Rejected: this is not supported by all architectures.
+
+Enarx client agent encrypts workload, sends to host agent. Host agent
+sends to Keep via a listening agent in the Keep. Keep decrypts, loads
+and runs. This maintains the host agent's status as a proxy between
+other components, and doesn't make further assumptions about routing
+which is made by 2. 
+Rejected:
+- the host agent should not need logic to provide functionality.
```
lkatalin: > the host agent should not need logic to provide functionality

Which logic? This sentence is a bit vague.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+- maintaining multiple versions is complex.
```
lkatalin: >  maintaining multiple versions is complex.

Multiple versions of what? What other versions are necessary alongside this / why?

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+## Rationale and alternatives
+
+Alternatives considered:
+
+Enarx client agent encrypts workload, sends to CPU/firmware via Enarx
+host agent. CPU/firmware decrypts workload, loads into Keep. This has
+the advantage that we use an existing network connection, and there's
+no need to establish a connection nor a session key between the Keep
+and the client agent.
+Rejected: this is not supported by all architectures.
```
lkatalin: > this is not supported by all architectures

Any examples?

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+- Since: 2020-01-10
+- Status Note: (has been subject to initial discussion and consensus)  
+- Supersedes: n/a
+- Start Date: 2020-02-16
+- Tags: architecture, webassembly
+
+## Summary
+
+This RFC describes the architecture adopted by Enarx for loading Wasm
+packages into a prepared Keep.
```
lkatalin:
```suggestion
packages into a prepared Keep using a secure channel. 
```

I think most of the discussion is about how to keep this loading process secure.

rfc#0001
```diff
@@ -0,0 +1,141 @@
...
+order for it to execute a workload, the Wasm (WebAssembly) package
+that comprises it (binary/binaries + data elements) need to be loaded
+into it.  This is sent to it by the Enarx client agent (ECA),
+encrypted under a session key established at Keep provisioning time,
+and communicated to the ECA - the workload provisioning encryption
+key (WPEK).  The Keep has an agent running which is listening on a
+well-known port, or on a port number previously established during
+the Keep preparation phase.  The ECA connects to this port, and
+establishes a connection to the Keep.  The Keep takes the workload,
+decrypts it, and runs it.
```
lkatalin:
```suggestion
decrypts it with the WPEK, and runs it.
```

lkatalin (changes requested):
I suggested a number of somewhat nitpicky changes here. Feel free to disregard anything I may have misunderstood. Most of the changes requested center around:

- Eliminating the use of "it" as a pronoun. The description is very technical and I think it should be painfully clear which components are doing what. Actually, most of these changes are about establishing painful clarity in all sentences.
- Being more detailed about where the WPEK comes from, which components are privvy to it, etc. There are some details omitted that would raise questions to an outside reader, IMO. The reader should know exactly what this key is so they understand why the channel is secure.

rfc#0001
MikeCamel: Hopefully improved.

rfc#0001
MikeCamel: Disambiguated.

rfc#0001
MikeCamel: @npmccallum - this was your point (which I captured, but failed to give an example).

rfc#0001
axelsimon: In this section also, i find "but since we may not have established any certificates in the Keep at this point, it cannot act as the server." confusing. TLS doesn't require a server and client, to my understanding it usually works using a Diffie-Hellman key exchange and therefore works peer to peer. What is hard here is not so much the encryption, but the authentication of the other party, which is where the Keep's lack of own certificates that it could use to identify itself becomes a problem.

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+One of the Enarx project's design principles is that all network
+and storage connections in and out of a Keep are encrypted.  In the
+case of network connections, this usually means that the session is
+encrypted using TLS (or dTLS), but since we may not have established
+any certificates in the Keep at this point, it cannot act as the
+server.  However, as the workload provisioning encryption key (WPEK)
+has already been communicated to the Keep, if the workload can be
+decrypted using this key, this assures the Keep that it is
+communicating with the ECA, or at least an entity in the same trust
+domain, so identity has been confirmed.
```
axelsimon:
```suggestion
domain, so tenant identity has been confirmed.
```

rfc#0001
axelsimon: I think it cares, but in a softer way than the tenant/ECA cares about the Keep: if a workload is encrypted with a WPEK, that's proof enough that it comes from whomever is allowed to send workloads.

Unencrypted workloads could be anything, so that's out of the question. Encrypted –and really, signed– workloads are the minimum for a Keep to ensure it's still "obeying the right master", so to speak.

rfc#0001
axelsimon: Personally, while i like the word "endpoint" i'd worry that "communication endpoint" is too vague in this case.

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+to verify their provenance.
+The Keep SHOULD listen on a well-known port (see "Unresolved
+questions").
+If a port has been pre-esablished, the Keep MUST listen on this port,
+and MAY also listen on the well-known port.
+The Keep MAY cache the workload for re-use, or transfer to another
+Keep.  If a Keep allows workloads to be used by another Keep, the
+workload MAY be encrypted under a new WPEK, established between the
+original Keep and new Keep.
+Keeps MAY be implemented to enable the capability to performing
```
axelsimon:
```suggestion
Keeps MAY be implemented to enable the capability to perform
```

rfc#0001
axelsimon: And so should we all!

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+enable caching.
+
+## Reference
+
+This RFC arose from this issue: [Workload (WASM binary) loading
+- connection termination](https://github.com/enarx/enarx/issues/140)
+
+## Drawbacks
+
+Why should we *not* do this?
```
axelsimon: I think this section, which comes from the template, should be filled in. Something like:
```suggestion
No drawbacks currently known. Furthermore, this mechanism is core functionality to Enarx.
```

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+In such implementations, a component trusted by the host system may
+handle key management and decryption/loading of the workload.
+
+## Unresolved questions
+
+1. What should the well-known port number be?  Note that a number
+above 1024 may make life a little easier (non-root binding).
+
+Out of scope:
+1. Keep preparation steps: requires RFC - issue #228.
```
axelsimon:
```suggestion
1. Keep preparation steps: requires RFC - issue #16.
```

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+1. Workload package format: requires RFC - issue #227.
```
axelsimon:
```suggestion
1. Workload package format: requires RFC - issue #17.
```

axelsimon (changes requested):

axelsimon:
Quick note, as per the RFC process (which is now live), this RFC should use the number of its PR, so it should be RFC 00002 :)

npmccallum (changes requested):
Please format the RFC according to the specification here:
https://github.com/enarx/rfcs/blob/master/contributing.md#how-to-propose-an-rfc

rfc#0001
axelsimon: I think this adds some useful clarification and is worth adding.

axelsimon (commented):
Looks good. Just one small comment to address, but otherwise it's there i think.

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+## Tutorial
+
+A Keep is the fundamental runtime component of Enarx: a TEE instance
++ Enarx runtime pieces, including WebAssembly and WASI layers.  In
+order for the Keep to execute a workload, the Wasm (WebAssembly) package
+that comprises the workload (binary/binaries + data elements) need to be loaded
+into the Keep.  This workload is sent to the Keep by the Enarx client agent
+(ECA), encrypted under a session key (the workload provisioning encryption
+key or WPEK) that is established at Keep provisioning time and
+communicated to the ECA.
```
lkatalin:
```suggestion
communicated between the ECA and the Keep using a Diffie-Hellman Key Exchange or similar.
```
I still think it would be better to word this more explicitly. "communicated to" is a bit vague as there is no subject and someone might infer that some third entity aside from the ECA and Keep is involved, or that the Keep is solely in charge of generating the key (which also isn't the case, at least the way we used it with SGX in the early demo).

rfc#0001
lkatalin: I'm okay with just ignoring this comment from earlier.

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+required (see below), proxying is possible.  This could, if required,
+be via an (untrusted) service on the host.
+
+One of the Enarx project's design principles is that all network
+and storage connections in and out of a Keep are encrypted.  In the
+case of network connections, this usually means that the session is
+encrypted using TLS (or dTLS), but since we may not have established
+any certificates in the Keep at this point, it cannot act as the
+server.  However, as the workload provisioning encryption key (WPEK)
+has already been communicated to the Keep, if the workload can be
```
lkatalin:
```suggestion
is already known to the Keep, if the workload can be
```

rfc#0001
lkatalin: @MikeCamel Thoughts?

rfc#0001
lkatalin: I'm okay with not having examples here.

lkatalin (changes requested):
This reads very well and my only outstanding comments are all related to the wording around the WPEK. The current RFC describes this as being "communicated to" both the ECA and (in a later paragraph) to the Keep; I think it would be clearer to say this is generated between them. This is implied by calling it a "session key," but then some of the other wording makes this relationship and the nature of the key and which components have access to it less clear.

connorkuehl (changes requested):
I wonder if the Enarx Client Agent needs to be broken up into separate entities. In my mind, the ECA is fulfilling a number of functions and it is hard to keep track of how it is accomplishing all of this. It seems to be the frontend utility for an end-user looking to deploy a workload but it also seems to forward itself to the host machine to negotiate the construction of a Keep (or many keeps). However, the way it is written now implies that the Keeps are _already present_ on the target machine and therefore are waiting on a socket for the client to tell it what to do. It is not clear what the workflow is if there is a new user with a new target machine to deploy on that has no keeps at all.

I'm wondering if we should decompose it into:

_The client_: a tool that someone can use on their local machine to manage their Enarx keeps + workloads.

_The daemon/agent (daemon sounds cooler):_ a long-running process on a target machine that will host Enarx keeps.

The user can use the client to dispatch commands to well-known hosts that are running the Enarx daemon. The Enarx daemon carries out the commands received from the client. This includes the construction and teardown of Enarx keeps.

With a separation of roles like this it is more clear how a binary makes its way into an Enarx keep regardless of where the Keep is in its life cycle and we can do away with the chicken-egg issue of assuming a Keep is already present and waiting on a socket on a target host.

Please rebase this branch against the most recent tip to squash all of these commits and add a `.md` extension to the file type so that GitHub can render this in HTML. 

It is helpful if you dismiss/re-request reviews once comments are addressed so it's clear that all pending thoughts have received closure and the RFC is ready for review again.

MikeCamel:
> I wonder if the Enarx Client Agent needs to be broken up into separate entities. In my mind, the ECA is fulfilling a number of functions and it is hard to keep track of how it is accomplishing all of this. It seems to be the frontend utility for an end-user looking to deploy a workload but it also seems to forward itself to the host machine to negotiate the construction of a Keep (or many keeps). However, the way it is written now implies that the Keeps are _already present_ on the target machine and therefore are waiting on a socket for the client to tell it what to do. It is not clear what the workflow is if there is a new user with a new target machine to deploy on that has no keeps at all.
> 
> I'm wondering if we should decompose it into:
> 
> _The client_: a tool that someone can use on their local machine to manage their Enarx keeps + workloads.
> 
> _The daemon/agent (daemon sounds cooler):_ a long-running process on a target machine that will host Enarx keeps.
> 
> The user can use the client to dispatch commands to well-known hosts that are running the Enarx daemon. The Enarx daemon carries out the commands received from the client. This includes the construction and teardown of Enarx keeps.
> 
> With a separation of roles like this it is more clear how a binary makes its way into an Enarx keep regardless of where the Keep is in its life cycle and we can do away with the chicken-egg issue of assuming a Keep is already present and waiting on a socket on a target host.

@connorkuehl - what you describe is the Enarx Host Agent.

connorkuehl:
> what you describe is the Enarx Host Agent.

@MikeCamel Indeed! That entity only appears under the rejected alternatives section and its function doesn't appear to have an explicit appearance in the proposed RFC :-)

MikeCamel:
I'll have a look.  I think you may be conflating some very different steps - which are described in detail in the latest pull request - but if it's not clear, then I need to elucidate.

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+- Tags: architecture, webassembly
+
+## Summary
+
+This RFC describes the architecture adopted by Enarx for loading Wasm
+packages into a prepared Keep using a secure channel. 
+
+## Motivation
+
+Once a Keep is prepared with microkernel, Wasm runtime, etc., the
```
connorkuehl: ~~microkernel~~ shim

Microkernel is true in the VM-based keeps but not the process-based keeps. Shim is a good term that fits both.

rfc#0001
```diff
@@ -0,0 +1,145 @@
+# WASM binary loading into an Enarx Keep
```
connorkuehl: ~~WASM~~ Wasm

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+that comprises the workload (binary/binaries + data elements) need to be loaded
+into the Keep.  This workload is sent to the Keep by the Enarx client agent
+(ECA), encrypted under a session key (the workload provisioning encryption
+key or WPEK) that is established at Keep provisioning time and
+communicated to the ECA.
+
+The Keep has an agent running which is listening on a
+well-known port, or on a port number previously established during
+the Keep preparation phase.  The ECA connects to this port, and
+establishes a connection to the Keep.  The Keep takes the workload,
```
connorkuehl: Should this say that it establishes a connection to the Enarx Host Agent or is Keep still correct here?

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+Some TEE implementations assume trust in the host system.  This is
+explicitly excluded as one of the Enarx project's design principles.
+In such implementations, a component trusted by the host system may
+handle key management and decryption/loading of the workload.
+
+## Unresolved questions
+
+1. What should the well-known port number be?  Note that a number
+above 1024 may make life a little easier (non-root binding).
+
```
connorkuehl: This strikes me as an implementation detail and probably something we want to be configurable by sysadmins who are setting up Enarx support on their compute resources. I don't think this question needs an answer that's more complicated than "whatever /etc/enarx/host.conf says the port is"

rfc#0001
```diff
@@ -0,0 +1,145 @@
...
+communicated to the ECA.
+
+The Keep has an agent running which is listening on a
+well-known port, or on a port number previously established during
+the Keep preparation phase.  The ECA connects to this port, and
+establishes a connection to the Keep.  The Keep takes the workload,
+decrypts it with the WPEK, and runs it.
+
+For this design to work, a route between the ECA and the Keep control
+plane needs to be available.  Note that this does not necessarily
```
connorkuehl: Similar question as above, should this be the route between the ECA and Enarx Host Agent?

connorkuehl (changes requested):

npmccallum:
@MikeCamel Although there have been numerous comments since, I haven't seen any updates to this RFC since Mar 16. Could you please update this RFC to address the comments so that we can do another round of review?

rfc#0001
MikeCamel: This is dealt with in detail in a separate RFC.  I've improved the wording somewhat.

rfc#0001
MikeCamel: Dealt with in a separate RFC.

rfc#0001
MikeCamel: I've made a number of changes to this section to address your points.

rfc#0001
MikeCamel: Took similar wording.

rfc#0001
MikeCamel: Agreed.  Added.

rfc#0001
MikeCamel: Point taken.  Anyone else?

rfc#0001
MikeCamel: Dealt with more specifically in other RFCs.

rfc#0001
MikeCamel: Fixed.

MikeCamel:
> Could you please update this RFC to address the comments so that we can do another round of review?

Done.

rfc#0001.md
```diff
@@ -0,0 +1,154 @@
...
+- Tags: architecture, webassembly
+
+## Summary
+
+This RFC describes the architecture adopted by Enarx for loading Wasm
+packages into a prepared Keep using a secure channel. 
+
+## Motivation
+
+Once a Keep is prepared with shim, Wasm runtime, etc., the
```
lkatalin, line 17:
```suggestion
Once a Keep is prepared with shim and Wasm runtime, the
```
The "etc." leaves me feeling confused about what else is in there.

rfc#0001.md
```diff
@@ -0,0 +1,154 @@
...
+tenant workload (Wasm binary + data) needs to be loaded into it.  A
+description of the loading sequence is required.
+
+## Tutorial
+
+A Keep is the fundamental runtime component of Enarx: a TEE instance
++ Enarx runtime pieces, including WebAssembly and WASI layers.  In
```
lkatalin, line 24: This is rendered strangely in markdown. "Enarx runtime pieces" becomes a bullet point.

rfc#0001.md
```diff
@@ -0,0 +1,154 @@
...
+design/implementation errors.  Any design of such functionality requires
+a new RFC (see "Future possibilities" below).
+
+The Keep MUST NOT accept unencrypted workloads, as there is no way
+to verify their provenance.
+The Keep SHOULD listen on a well-known port (see "Unresolved
+questions").
+If a port has been pre-established, between the Enarx Client Agent and
+the Keep, then the Keep MUST listen on this port,
+and MAY also listen on the well-known port.
```
lkatalin, line 83: Why would the Keep listen on both ports?

rfc#0001.md
```diff
@@ -0,0 +1,154 @@
...
+Enabling of the capability for a Keep to perform caching MUST be at
+the discretion of the ECA or entities within the ECA's trust domain.
+Enabling of the capability for a Keep to perform caching MUST NOT
+be at the discretion of the host.  The host MUST NOT be able to
+enable caching.
+
+## Reference
+
+This RFC arose from this issue: [Workload (WASM binary) loading
+- connection termination](https://github.com/enarx/enarx/issues/140)
```
lkatalin, line 99: This also renders strangely in markdown.

rfc#0001.md
```diff
@@ -0,0 +1,154 @@
...
+the advantage that we use an existing network connection, and there's
+no need to establish a connection nor a session key between the Keep
+and the client agent.
+Rejected: this is not supported by all architectures.
+
+Enarx client agent encrypts workload, sends to host agent. Host agent
+sends to Keep via a listening agent in the Keep. Keep decrypts, loads
+and runs. This maintains the host agent's status as a proxy between
+other components, and doesn't make further assumptions about routing
+which is made by 2.
```
lkatalin, line 120: What's "2"?

rfc#0001.md
```diff
@@ -0,0 +1,154 @@
...
+Enarx client agent encrypts workload, sends to host agent. Host agent
+sends to Keep via a listening agent in the Keep. Keep decrypts, loads
+and runs. This maintains the host agent's status as a proxy between
+other components, and doesn't make further assumptions about routing
+which is made by 2. 
+Rejected:
+- the host agent should act as an untrusted proxy wherever possible,
+and not include process logic.
+- maintaining multiple versions of host agents is complex from
+a deployment point of view.
```
lkatalin, line 125: I don't actually understand why process logic is required for the host agent to be a proxy in this way, or why multiple versions of the host agent would be necessary with this design.

lkatalin (changes requested):
I actually think this is pretty close. There are a few formatting errors and I had a question about one of the rejected proposals.

PR #24, opened 2020-03-30
Fix RFC name syntax checker

npmccallum:
Also, add the ability to have some RFC-associated files (such as images).


npmccallum:
@mbestavros I believe your RFC should pass under this test.

mbestavros (approved):
LGTM!

PR #27, opened 2020-05-07
Detail a sprint process for the Enarx project

mbestavros:
You can view the formatted version [here](https://github.com/mbestavros/rfcs/blob/sprints/00003-enarx-sprint-framework/README.md).

Please provide feedback -- discussion on this is important!


MikeCamel (changes requested):
This is a good start as a discussion and rationale document, but what I believe we need is a definition and description of the process.  This should be more formal:

- few first person plural pronouns ("we"/"our"), mainly third person singular ("it");
- fewer "should", more "will";
- steps laid out in (preferably numbered) bullet form for each of the meetings.

Much of the existing text feels like a separate discussion document or appendix.

Crucially missing is a list of roles and responsibilities.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
+# Agile Sprint Process for Enarx
+- Authors: [Mark Bestavros](mbestavr@redhat.com) -- email is optional
```
connorkuehl: ~~email is optional~~

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-05-06 (date you submit your PR)
```
connorkuehl: ~~(date you submit your PR)~~

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Status Note: Concept brainstormed, awaiting implementation
+- Supersedes: N/A
+- Start Date: 2020-05-01 (date you started working on this idea)
```
connorkuehl: ~~(date you started working on this idea)~~

npmccallum (changes requested):
@mbestavros This is a good start.

1. Please wrap the text at <80 characters. This makes inline review easier.

2. I'd like to start with a single backlog and then split it up as it makes sense in the future.

3. This process has a variety of gaps. For example, when does an issue get marked as done? When does a PR get merged? We don't currently have a formal process here and need to grow one.

4. I think this RFC would be significantly helped by a process flowchart. Try [this](https://dreampuf.github.io/GraphvizOnline).

5. There is no discussion of the following: user stories, epics, teams, milestones and releases (I might be missing some). I think it would be helpful to at least mention them, even if that is "not at this time."

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Triage
+
+By default, we shouldn't do anything when an issue is first submitted. The lack of metadata will indicate it needs to be triaged. (If desired, we could post a quick welcome comment if they're not part of the core team, but that's optional.)
+
+Every standup, we'll look at all issues that are not part of a project and not labeled. This ensures that new issues always get a human response within 24 working hours, and that all issues are organized properly into the backlog -- or, if needed, promoted to the current sprint. It also allows the team to discuss all new issues as they come up.
+
+### Backlog
+
+As our project scales, it’s easy to imagine a single unified backlog becoming incredibly cumbersome and hard to manage. Instead, we should adapt our current approach to area-specific boards, which effectively divides our backlog into categories (SGX, SEV,  infrastructure, etc.).
```
connorkuehl: My first reaction as I read this paragraph was to agree. However, I think there might be some value to consolidating our boards into a single board. The primary reason is that when we're sprinting, I'd argue that the single most important unit of work is our active sprint (comprised of issues that should arguably sum to a reasonable amount of work within a workweek per engineer). By having just one board, we can capture the contents of the sprint with one column on one board, rather than having this fragmented across component-level boards. I think we can also accomplish this with GitHub milestones.

In any case, I don't think smaller boards buy us anything. Even if the rate at which we accumulate work vastly outpaces the work completed per sprint, I think we'd benefit more from disciplined grooming of the backlog and decomposition of tasks than fragmenting the backlog into component-sized boards. We will need to advocate ruthlessly for understanding the priority of individual issues and how they fit in with our roadmap moving forward. Regardless of the number of boards, the size of the backlog is the size of the backlog, regardless of if it's divided by 1 board or N-boards.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+Here’s how the columns should look:
+
+- Assigned: this indicates a task that is currently being worked on as part of the sprint. Newly-added issues land here by default (automatable with stock Github).
+- Ready for review: where pull requests land by default (automatable with stock Github); indicates it’s ready to be reviewed by someone. Pull requests should move back here from Changes Requested if a fix has been applied and it’s ready for a re-review (not in stock Github, but automatable with Actions).
+- Changes Requested: PRs move here when a review is submitted that requests changes (automatable with stock Github).
+- Approved: PRs that have been approved go here, pending merge (automatable with stock Github).
+- Done: PRs that have been merged, and their associated issues, go here (automatable with stock Github).
+
+A few things to consider:
+- Advantage: having project board be separate from backlog means we can promote issues to be part of the current sprint whenever we want to, and they can skip the backlog entirely if that’s desired.
```
connorkuehl: I think we should be cautious with a mechanism like this. If we abuse it then it undermines the efforts that go into sprint planning.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+This is meant to be a time to consider what work the team wants to accomplish in the next two-week sprint and assign relevant tasks to people. It should happen on Mondays and take about an hour, following the sprint retrospective (detailed later).
+
+It should start with a brief discussion about the team’s goals. This should be an open, but high-level, discussion that will help frame the kinds of issues we should prioritize. It shouldn’t take long -- no more than 15-20 minutes.
+
+With the context of that discussion in mind, we’ll move to backlog review. As detailed above, our backlogs are contained in individual boards, and we’ll run through each one as part of the planning process.
+
+For each board:
+- Briefly run through the Backlog column to decide if there are any issues we want to promote to "Tracking" status.
+- Run through all items in "Tracking" and decide which ones are highest priority, and who will work on them. They will become the issue’s assignee.
```
connorkuehl: I think it's important that we come up with some agreeable "unit" of work and estimate how many units of work comprise a "full" sprint for an engineer.

Furthermore, when grooming or picking tasks we should have some consensus based on the estimated units of work that issue will take. Just to be clear, this is not a do-or-die deadline (and it's most certainly not to be used _against_ anyone during retro), but this is a tool for helping us not overload someone with work. Estimating scope of most non-trivial issues is exceptionally hard, especially since we have such a component-based group, but I think it's still worthwhile. These metrics may be useful during the retrospective. If the unit is agreed on as a team, that ties in to the fact that we sprint as a team and we ultimately move code as a team. If issues are left unfinished due to scope or difficulty, then at least we can share that learning experience as a team.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+First, if there are any new people on the call, we'll welcome them and allow them to ask questions or have discussions as needed.
+
+Next, we’ll have a look at all untriaged issues -- any that do not have a label, and are not part of a project. This will allow the whole team to discuss them and ensure they get moved to the appropriate backlog board and labeled correctly. This shouldn’t take more than 15 minutes, though that’s a soft limit -- if a good conversation is happening, it shouldn’t be cut off.
+
+Finally, if there's time, we'll dedicate some time to a quick review of our Current Sprint board. This will serve as an opportunity for developers to elaborate on any blockers and ask for reviews or re-reviews as needed. We should strive to do this every day if possible, but if prior discussion makes that infeasible, we don't _have_ to do it (there's always tomorrow). Project discussion is more important.
+
+### Sprint Retrospective
+Our final meeting is the sprint retrospective. This should be a quick half-hour meeting that takes the place of Monday’s standup where we discuss:
```
connorkuehl: I think we should err on the side of caution and block off 60 minutes for the retrospective.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- What we accomplished in the last sprint (perhaps with a demo, if appropriate)
+- What didn’t get done, and why (blockers? More complex than anticipated? etc)
+
+If there is leftover work on our Active Sprint board, we should decide whether we should continue working on it or demote it to a backlog board. Then, we move onto the next sprint and do it all over again.
```
connorkuehl: We should discuss why it didn't get completed. This might happen naturally as whoever is working on it explains blockers throughout the week, but broadly speaking understanding the consequences of our planning will help us plan better moving forward. There's a good chance we underestimated the difficulty, or perhaps there was a lot of scaffolding work that needed to fall into place before this one did that we didn't foresee. In both cases, we should regroom the issue and further decompose it if necessary.

concepts/00001-enarx-sprint-framework/README.md
npmccallum: @connorkuehl Agreed. However, I asked for this explicitly because I want a way to capture drive-by contributions. I don't want a new contributor to have to wait 2 weeks for a PR just because they missed the planning meeting by 5 minutes.

concepts/00001-enarx-sprint-framework/README.md
lkatalin: I just checked, and we are able to filter cards by labels while on a board, so this could help us stay organized if we do decide to have just one backlog instead of multiple backlogs.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- SGX
+- WASM
+
+A few things to consider:
+- Downside: issues that don’t fit squarely into a category don’t have a place to go. Maybe a "Miscellaneous" project?
+- Automation opportunity: we could automatically move issues from Tracking to Active when they are assigned to someone. Automatable with Actions.
+
+### Active Sprint Board
+This board is where the team will track progress on current sprint tasks. **It should hold no backlog;** if an issue or PR is on this board, it is being actively worked on.
+
```
lkatalin: Moving issues and PRs between boards is easily done, right? I'm assuming so but I've never tried it myself. Ease of doing this could inform whether we want the component-specific backlog boards.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+Here’s how the columns should look:
+
+- Assigned: this indicates a task that is currently being worked on as part of the sprint. Newly-added issues land here by default (automatable with stock Github).
+- Ready for review: where pull requests land by default (automatable with stock Github); indicates it’s ready to be reviewed by someone. Pull requests should move back here from Changes Requested if a fix has been applied and it’s ready for a re-review (not in stock Github, but automatable with Actions).
+- Changes Requested: PRs move here when a review is submitted that requests changes (automatable with stock Github).
+- Approved: PRs that have been approved go here, pending merge (automatable with stock Github).
+- Done: PRs that have been merged, and their associated issues, go here (automatable with stock Github).
```
lkatalin: Does "done" on the sprint board automatically move issues to "done" on the backlog boards? Are the issues duplicates of each other? I'm not 100% on how backlog and sprint stay in sync since we don't have pointers on GitHub boards....

concepts/00001-enarx-sprint-framework/README.md
lkatalin: I also have it come up pretty often that something I'm working on (or something that comes up within @npmccallum 's work) spawns a new issue that gets promoted to my immediate workload... so not being able to do that seems strange? Cumbersome? I can think of sprint-compliant ways to do this sort of thing, but we should probably talk about what we want that to look like.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+A few things to consider:
+- Advantage: having project board be separate from backlog means we can promote issues to be part of the current sprint whenever we want to, and they can skip the backlog entirely if that’s desired.
+- Automation opportunity: If pull requests have an associated issue, don’t put the PR on the board -- instead, move the associated issue to the board (if not on it already) and treat the issue like the pull request, subject to the same movement rules as specified above. This would prevent us from having to look at the same thing twice, while still preserving a direct link to the PR in the context of its issue (since Github now displays linked PRs as a dropdown on an issue).
```
lkatalin: > If pull requests have an associated issue

Would they ever not?

lkatalin: I'm curious if this might address my question about duplicate issues between sprint / component boards. So the issue goes on the sprint board and the PR goes on the component board?

concepts/00001-enarx-sprint-framework/README.md
lkatalin: I agree with @connorkuehl , I think what a "unit" of work is and what each task is worth should be considered carefully. I will admit up front that I have no idea how long a task might take more than half the time, because I mostly do things I've never done before. Sometimes a task I think will take a day takes three weeks, and sometimes (rarely) a task I think will take two weeks takes an afternoon. Others can probably help out here in estimating, but we also have to keep in mind that something that might take a senior engineer two days could take a more junior engineer two weeks or an intern a month. IMO this is the biggest challenge with moving to a "sprint" structure since so much of our project is uncharted territory and new problems come up en route to even straightforward-seeming issues.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+With the context of that discussion in mind, we’ll move to backlog review. As detailed above, our backlogs are contained in individual boards, and we’ll run through each one as part of the planning process.
+
+For each board:
+- Briefly run through the Backlog column to decide if there are any issues we want to promote to "Tracking" status.
+- Run through all items in "Tracking" and decide which ones are highest priority, and who will work on them. They will become the issue’s assignee.
+
+From there, automation will aid the process -- any newly assigned issues will:
+- be moved to the Active column in the backlog board
+- automatically added to the Assigned column of the Active Sprint board
```
lkatalin: Oh, so it does sound like there _are_ duplicate issues?

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Daily Standup
+Our daily standups will take place every day we don’t have a longer meeting (such as sprint planning/retro) for half an hour. This is meant to be a quick check-in to ensure newly-opened issues get quick responses and blockers get resolved.
+
+First, if there are any new people on the call, we'll welcome them and allow them to ask questions or have discussions as needed.
+
+Next, we’ll have a look at all untriaged issues -- any that do not have a label, and are not part of a project. This will allow the whole team to discuss them and ensure they get moved to the appropriate backlog board and labeled correctly. This shouldn’t take more than 15 minutes, though that’s a soft limit -- if a good conversation is happening, it shouldn’t be cut off.
+
+Finally, if there's time, we'll dedicate some time to a quick review of our Current Sprint board. This will serve as an opportunity for developers to elaborate on any blockers and ask for reviews or re-reviews as needed. We should strive to do this every day if possible, but if prior discussion makes that infeasible, we don't _have_ to do it (there's always tomorrow). Project discussion is more important.
```
lkatalin: Maybe it's just me, but it seems like a daily discussion of blockers / opportunity to ask questions and report progress is the biggest value of our current stand-up. Do we actively want this to change? If so, that could be fine (we still have the chat for that sort of thing) but I just wanted to point out that this description seems very different from what we're doing currently.

concepts/00001-enarx-sprint-framework/README.md
lkatalin: A 60 minute retrospective could make sense. If we do this and the planning meeting is the same day, I would strongly vote to have them back-to-back so that if the retrospective only takes 25 minutes we can immediately transition into planning, instead of having an awkward gap between the two and having two hours blocked off that we don't really need but where it's hard to get work done.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+For the first few sprints we do, it may be worth setting some time aside for process feedback: how well is our agile approach working? How can we improve it? etc.
+
+## Drawbacks
+
+- The team will need to attend a few more meetings. The daily standups, as currently designed, lose their focus on individual areas of the project per day; this was useful in case an external contributor was interested in one specific area of the project. However, we hope the added free-form discussion time will allow for relevant questions to be asked and talked about anyway.
+
+- Implementing some of the automation exactly as specified may end up being a challenge. There's further discussion of this below.
+
+- Bisecting our backlog into different projects has its benefits, but it also precludes a nice, unified view of what needs to be done on the project. This would be addressable with a `backlog` label if we considered that a large-enough problem.
```
lkatalin: Yeah, I think using labels for more of the organization and dropping the different boards could make sense - currently a lot of our labels are redundant with the "projects." I don't have a strong opinion on this, though.

lkatalin (commented):
This seems well thought out! I'm commenting instead of approving or requesting changes, since I think there will be more to discuss as a group.

My biggest concern with the sprint process in general:
Estimating units of work and how long tasks will take seems like it will need a wide margin for error, since we are doing things that are difficult and novel.

My biggest concern with the specific proposed plan:
I'm not sure how the backlog board and sprint board issues will relate to each other, and I can't think of a way to tie them together that isn't awkward. Will we have two duplicate issues, one for each board? How will we ensure they both reflect updates? Or will we have one issue that moves from backlog -> sprint -> backlog? None of these options seems very smooth when I think about it.

Otherwise this generally seems workable. More comments on the doc.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Start Date: 2020-05-01
+- Tags: process, sprint, concept
+
+## Summary
+
+Enarx has been using its daily standup and board organization for a while. It's worked for us as a small team, but as we grow, we anticipate it becoming unsustainable. It also doesn't do a good job capturing the current thrust of our work. This RFC proposes a replacement process, designed around the agile sprint methodology.
+
+## Motivation
+
+The Enarx project needs to organize its work better, and in a more sustainable way than before. We need to do a better job planning for medium-term work and goals, and in a way that scales more effectively than our prior process.
```
axelsimon:
```suggestion
The Enarx project needs to organize its work better, and in a more sustainable way than before. The project needs to do a better job planning for medium-term work and goals, and in a way that scales more effectively than our prior process. As we anticipate the project's separation into sub-teams working on specific aspects of Enarx to grow, we need to organise work in such a way that anyone can easily know what each team is working on right now, what are its priorities, and which team member is assigned to which task.
```

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Tutorial
+
+Before we start thinking about how our sprints will work, we should lay the groundwork for how our backlog and sprint boards should be organized.
+
+### Triage
+
+By default, we shouldn't do anything when an issue is first submitted. The lack of metadata will indicate it needs to be triaged. (If desired, we could post a quick welcome comment if they're not part of the core team, but that's optional.)
+
+Every standup, we'll look at all issues that are not part of a project and not labeled. This ensures that new issues always get a human response within 24 working hours, and that all issues are organized properly into the backlog -- or, if needed, promoted to the current sprint. It also allows the team to discuss all new issues as they come up.
```
axelsimon: I'm not sure we should do this, it sounds quite time consuming: actually qualifying an issue and ensuring everyone shares a common understanding of it takes time. Most issues don't need an answer within 24 hours. If they truly are urgent, they could be labelled "urgent". If they are security-related, there's another process for that. Lastly, any remaining issue *can* still be addressed shortly, if nothing else prevents it.

The usual process in agile sprints is to have a backlog grooming session which is used to get a set of stories ready to a point where they can be simply picked up for the next sprint.
This doesn't mean we can't discuss new issues as they come up, but usually the full conversation and especially the alignment of understanding and expectations would be kept for the backlog grooming session.

Lastly, "promoted to the current sprint" is something i would advise against. The point of a sprint is that it protects you against outside interference : the team decides what it will work on during the sprint and gets to delivering just that. Sprints are short, so anything else stakeholders want can actually wait until the next one, 95% of the time. You want to avoid setting a precedent of allowing issues opened halfway through the sprint to be accepted in the sprint, disrupting it.
There can still be exceptions, of course (as suggested in paragraph one) but in any case, they should be done on a one-in-one-out principle, as the team has already planned the sprint to do just as much as it can, no more, no less.

concepts/00001-enarx-sprint-framework/README.md
axelsimon: I agree with @connorkuehl here. One issue would sub-boards also arises when you need to work on something that affects multiple components at the same time.

That's not to say they can't be useful, as filtering out stuff to just see what you need to see is clearly useful at times.

Sub-boards as just a filtered "main board" seem like a good way to look at this to me.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Backlog: for issues we want to tackle eventually, but not a high priority
+- Tracking: for issues that are high-priority, and that we want to take action on sooner rather than later. All new project issues should land here by default to ensure they don’t get lost (automatable with stock Github).
+- Active: for issues that are actively being worked on as part of the current sprint.
+- Done: when things are done, they go here (automatable with stock Github). Mostly to ensure the “Active” column stays organized.
+
+In addition to inherently categorizing our backlog, this approach allows new "categories" to be moved around or created as needed -- and, if we decide to use organization-level projects for our sprints, allows the backlog to be bisected across repositories if wanted. This will support modularization as the project grows.
+
+To start, I propose the following backlog projects:
+
+- Documentation
```
axelsimon: Ideally, documentation would not be a separate thing but built into each story/issue (and an agree-upon condition for considering something "done"), but that can also be considered to be a agile purist position :)

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+Each area-specific backlog will be housed in a Github project with the following columns:
+
+- Backlog: for issues we want to tackle eventually, but not a high priority
+- Tracking: for issues that are high-priority, and that we want to take action on sooner rather than later. All new project issues should land here by default to ensure they don’t get lost (automatable with stock Github).
+- Active: for issues that are actively being worked on as part of the current sprint.
+- Done: when things are done, they go here (automatable with stock Github). Mostly to ensure the “Active” column stays organized.
+
+In addition to inherently categorizing our backlog, this approach allows new "categories" to be moved around or created as needed -- and, if we decide to use organization-level projects for our sprints, allows the backlog to be bisected across repositories if wanted. This will support modularization as the project grows.
```
axelsimon:
```
This will support modularization as the project grows.
```
:+1:

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+To start, I propose the following backlog projects:
+
+- Documentation
+- Infrastructure
+- SEV
+- SGX
+- WASM
+
+A few things to consider:
+- Downside: issues that don’t fit squarely into a category don’t have a place to go. Maybe a "Miscellaneous" project?
```
axelsimon: This is where just treating each sub-board as a filtered view of the main, global board helps: any issue not specific to a sub-project is a general, Enarx issue.

concepts/00001-enarx-sprint-framework/README.md
axelsimon: At least for issues, it's just matter of picking which board to assign it to, as long as the issue and board you want are part of the same repo. Cross-repo transfer is more involved (but perfectly doable).

concepts/00001-enarx-sprint-framework/README.md
axelsimon: > Agreed. However, I asked for this explicitly because I want a way to capture drive-by contributions. I don't want a new contributor to have to wait 2 weeks for a PR just because they missed the planning meeting by 5 minutes.

This makes sense, but i think drive-by contributors should not be considered to be "part of the team", and as such exempt from following sprints.

@lkatalin According to agile doctrine, this would suggest that the original issue was either not groomed properly, or not broken down into the right subtasks during sprint planning. The spawned issue you run into would likely have emerged as the issue-owner tries to explain the issue to the rest of the team and everyone reaches a common understanding (backlog grooming) or as the issue is broken down into immediately actionable tasks (sprint planning).

This is all a bit theoretical of course, because there's always stuff that just wasn't foreseeable. But then again, the concept of spikes, or stories to explore something that isn't well understood enough to be done as a normal user story exists for this case. :smiley:

axelsimon (commented):
Leaving a first lot of comments, will get back to this later. It looks overall very good and close to being usable!

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: A "Sprint board" is going to be important - I see this noted below.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: No, it's not purist: we need documentation to be part of each story/issue.  But there are also non-component documentation tasks which may need to be considered separately.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: There are also particular roles to be considered here: it's not just a free-for-all discussion.  It needs to be led and coordinated.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: Agree - I'd expect individual reports to happen before untriaged issues discussion.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: "We" is dangerous here.  Again - there needs to be a discussion of roles here.  Until we have the roles described, we're going to struggle with what happens in each type of meeting.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Prior art
+
+Our current process of topic-specific boards largely informs this proposal's approach to the backlog. In fact, this proposal could be considered an extension or adaptation of our current process in some ways.
+
+And, of course, the agile sprint methodology is widely employed just about everywhere in software; see [Atlassian](https://www.atlassian.com/agile/scrum/sprints) and [Trello](https://blog.trello.com/beginners-guide-scrum-and-agile-project-management) for good examples. And, of course, Red Hat [itself](https://developers.redhat.com/blog/2017/06/29/agile-software-development-the-red-hat-way/) uses it.
+
+## Unresolved questions
+
+The most significant unresolved question does not involve process, but rather the automation we hope to pair with it: whether the Active Sprint board will live inside the main Enarx repository, or as an all-encompassing Enarx organization board.
```
MikeCamel: Disagree: for me, the most signficant unresolved question revolves around role definitions and the responsibility of each role in the various steps in the process.

concepts/00001-enarx-sprint-framework/README.md
axelsimon: Good point.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Active Sprint Board
+This board is where the team will track progress on current sprint tasks. **It should hold no backlog;** if an issue or PR is on this board, it is being actively worked on.
+
+Here’s how the columns should look:
+
+- Assigned: this indicates a task that is currently being worked on as part of the sprint. Newly-added issues land here by default (automatable with stock Github).
+- Ready for review: where pull requests land by default (automatable with stock Github); indicates it’s ready to be reviewed by someone. Pull requests should move back here from Changes Requested if a fix has been applied and it’s ready for a re-review (not in stock Github, but automatable with Actions).
+- Changes Requested: PRs move here when a review is submitted that requests changes (automatable with stock Github).
+- Approved: PRs that have been approved go here, pending merge (automatable with stock Github).
```
axelsimon: I'm not sure what the Approved state gives us, i might be missing something.
Is there anything more to this state than describing PRs where everything is good, but simply haven't been merged yet? Would there be a reason to hold back and wait on merging said PRs if everything is good with them?

If not, it sounds like we can go directly to done, by simply merging any PR in "ready to review" which doesn't require any further changes.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Automation opportunity: If pull requests have an associated issue, don’t put the PR on the board -- instead, move the associated issue to the board (if not on it already) and treat the issue like the pull request, subject to the same movement rules as specified above. This would prevent us from having to look at the same thing twice, while still preserving a direct link to the PR in the context of its issue (since Github now displays linked PRs as a dropdown on an issue).
+
+### Sprint Planning
+With our backlog and sprint boards established, we can flesh out our first meeting: sprint planning.
+
+This is meant to be a time to consider what work the team wants to accomplish in the next two-week sprint and assign relevant tasks to people. It should happen on Mondays and take about an hour, following the sprint retrospective (detailed later).
+
+It should start with a brief discussion about the team’s goals. This should be an open, but high-level, discussion that will help frame the kinds of issues we should prioritize. It shouldn’t take long -- no more than 15-20 minutes.
+
```
axelsimon: Very much agreed, aligning on what we want to achieve as a team over the coming sprint is both helpful for individual work, as you know why you are doing what you are doing, but also creates a better sense of having a common goal and puts said individual work in context.

concepts/00001-enarx-sprint-framework/README.md
axelsimon: Being able to flag any blockers and generate the opportunity to address them at some time after the meeting is one of the most valuable aspects of the daily stand-up, i think. So I agree with @lkatalin and @MikeCamel above.
To be specific, i think this aspect is more important than "what i have done, what i'm doing next" (although they are also valuable).

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Changes Requested: PRs move here when a review is submitted that requests changes (automatable with stock Github).
+- Approved: PRs that have been approved go here, pending merge (automatable with stock Github).
+- Done: PRs that have been merged, and their associated issues, go here (automatable with stock Github).
+
+A few things to consider:
+- Advantage: having project board be separate from backlog means we can promote issues to be part of the current sprint whenever we want to, and they can skip the backlog entirely if that’s desired.
+- Automation opportunity: If pull requests have an associated issue, don’t put the PR on the board -- instead, move the associated issue to the board (if not on it already) and treat the issue like the pull request, subject to the same movement rules as specified above. This would prevent us from having to look at the same thing twice, while still preserving a direct link to the PR in the context of its issue (since Github now displays linked PRs as a dropdown on an issue).
+
+### Sprint Planning
+With our backlog and sprint boards established, we can flesh out our first meeting: sprint planning.
```
axelsimon: In my experience of agile methodologies, the first meeting, which is a prerequisite to sprint planning, is backlog grooming.
The idea is:
1. backlog grooming: the team looks through the backlog of unrefined issues, and selects the next ones to groom. Any new issue that someone brings to the table as important is a good candidate, as are issues that are just the logical next step. The person who opened the issue explains it to the team, and a conversation regarding its importance, difficulty and, importantly, the "definition of done" occurs. The team then votes to assign "story points", which are arbitrary but serve as a way to compare team members' understanding and reveals any gap in comprehension. When a common value is reached, it's recorded on the story/issue and the issue is now one that can be picked for a sprint, during sprint planning.
- sprint planning: having properly groomed the stories/issues in the previous meeting, the team can simply pick a bunch of stories/issues for the sprint according to two things:
  - the goal of the sprint
  - the total number of points the stories/issues add up to
Sprint planning is also where, theoretically, stories get divided into smaller, immediately actionable tasks. In practise, thinking ahead to these tasks helps realise the complexity of the issue and therefore its point value (this should also help answer some of @lkatalin's questions!).

So normally, the sprint planning is a lot easier and faster, than the grooming.

So i suppose my question is: when is grooming of issues supposed to be done? I think your suggestion is during triage, so a little bit every day. I have no experience of doing that, and as mentioned in my comment re: triage, i have doubts regarding how it would work out given how time consuming it is.

axelsimon: In my experience of agile methodologies, the first meeting, which is a prerequisite to sprint planning, is backlog grooming.
The idea is:
1. backlog grooming: the team looks through the backlog of unrefined issues, and selects the next ones to groom. Any new issue that someone brings to the table as important is a good candidate, as are issues that are just the logical next step. The person who opened the issue explains it to the team, and a conversation regarding its importance, difficulty and, importantly, the "definition of done" occurs. The team then votes to assign "story points", which are arbitrary but serve as a way to compare team members' understanding and reveals any gap in comprehension. When a common value is reached, it's recorded on the story/issue and the issue is now one that can be picked for a sprint, during sprint planning.
2. sprint planning: having properly groomed the stories/issues in the previous meeting, the team can simply pick a bunch of stories/issues for the sprint according to two things:
  - the goal of the sprint
  - the total number of points the stories/issues add up to
Sprint planning is also where, theoretically, stories get divided into smaller, immediately actionable tasks. In practise, thinking ahead to these tasks helps realise the complexity of the issue and therefore its point value (this should also help answer some of @lkatalin's questions!).

So normally, the sprint planning is a lot easier and faster, than the grooming.

So i suppose my question is: when is grooming of issues supposed to be done? I think your suggestion is during triage, so a little bit every day. I have no experience of doing that, and as mentioned in my comment re: triage, i have doubts regarding how it would work out given how time consuming it is.

axelsimon: In my experience of agile methodologies, the first meeting, which is a prerequisite to sprint planning, is backlog grooming.
The idea is:
1. backlog grooming: the team looks through the backlog of unrefined issues, and selects the next ones to groom. Any new issue that someone brings to the table as important is a good candidate, as are issues that are just the logical next step. The person who opened the issue explains it to the team, and a conversation regarding its importance, difficulty and, importantly, the "definition of done" occurs. The team then votes to assign "story points", which are arbitrary but serve as a way to compare team members' understanding and reveals any gap in comprehension. When a common value is reached, it's recorded on the story/issue and the issue is now one that can be picked for a sprint, during sprint planning.
2. sprint planning: having properly groomed the stories/issues in the previous meeting, the team can simply pick a bunch of stories/issues for the sprint according to two things:
  - the goal of the sprint
  - the total number of points the stories/issues add up to
Sprint planning is also where, theoretically, stories get divided into smaller, immediately actionable tasks. In practise, thinking ahead to these tasks helps realise the complexity of the issue and therefore its point value (this should also help answer some of @lkatalin's questions!).

So normally, the sprint planning is a lot easier and faster, than the grooming.

So i suppose my question is: when is grooming of issues supposed to be done? I think your suggestion is during triage, so a little bit every day. I have no experience of doing that, and as mentioned in my comment re: triage, i have doubts regarding how it would work out given how time consuming it is.

concepts/00001-enarx-sprint-framework/README.md
axelsimon: The unit of work is supposed to stem from the "complexity points", AFAIK. See comment above regarding backlog grooming vs. sprint planning on this.

axelsimon (commented):
More comments. Sorry for leaving them in multiple batches.

concepts/00001-enarx-sprint-framework/README.md
blazebissar: Agree with the above comments: without clear roles and responsibilities it is easy to repeat the story of [Everybody, Anybody, Somebody, and Nobody](https://www.lollydaskal.com/leadership/story-everybody-somebody-anybody-nobody/) or just talk in circles without reaching an output. I've seen that meetings tend to be ineffective without understood inputs, agendas/roles, and outputs. In this case, the outputs should also include how to improve the process itself (as mentioned below), how to address the issues experienced during the sprint, and the personnel responsible for each output task.

concepts/00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,158 @@
...
+- What we accomplished in the last sprint (perhaps with a demo, if appropriate)
+- What didn’t get done, and why (blockers? More complex than anticipated? etc)
+
+If there is leftover work on our Active Sprint board, we should decide whether we should continue working on it or demote it to a backlog board. Then, we move onto the next sprint and do it all over again.
+
+For the first few sprints we do, it may be worth setting some time aside for process feedback: how well is our agile approach working? How can we improve it? etc.
+
+## Drawbacks
+
+- The team will need to attend a few more meetings. The daily standups, as currently designed, lose their focus on individual areas of the project per day; this was useful in case an external contributor was interested in one specific area of the project. However, we hope the added free-form discussion time will allow for relevant questions to be asked and talked about anyway.
```
blazebissar: As a newcomer, I actually think the free-form discussion is more valuable than the area-focus approach. The responsiveness and availability of the core team and the good use of chat make it easy to engage about a specific area as well.

Additionally, newcomer and external contributers are not present at most meetings and are not the majority of participants, so I think it is reasonable to structure meetings to maximally enable the core team and make adjustments as needed for new members when they do show up.

concepts/00001-enarx-sprint-framework/README.md
m-dombrowski: Automation was my first flag early on and I see it addressed later in the text. I appreciate the breakdown of concepts in the document. Please forgive my ignorance, I must ask how much this alternate process will streamline efforts. More to the point, does this process reduce or increase overall administrative overhead in addition to engineer related tasks? If it does streamline, is there an associated risk of "paralysis by analysis?"

The document indicates a notable level of uncertainty regarding the automation of this process and a less centralized use of a platform. Is there an understood exit strategy if the transition fails (eg automation) and the project needs to revert to existing project management methods, in whole or in part?

concepts/00001-enarx-sprint-framework/README.md
npmccallum: @axelsimon My concern here is that new contributors who file issues or PRs get prompt engagement from the team. Let's say new contributor Bob files a PR immediately after the grooming meeting. Should he wait 2 weeks for his PR to get merged? Agile wasn't designed with open source in mind. So I'm looking for a tweak to agile retain open source distinctives.

concepts/00001-enarx-sprint-framework/README.md
lkatalin: @axelsimon Yeah, I think this partially speaks to my (mild) skepticism that some of our workflow can entirely fit into the Platonic ideal "sprint" because it's all new ground and we often don't know what the sub-tasks or blockers are until we try a few different things. But it's also possible that we can be more exact about sub-tasks than we currently are. Only time will tell. I am okay with trying to conform to the sprint and seeing how it goes, and adapting it as required.

concepts/00001-enarx-sprint-framework/README.md
lkatalin: Thanks @axelsimon :) these comments are useful to read.

concepts/00001-enarx-sprint-framework/README.md
npmccallum: I am noting consensus on a single sprint board.

concepts/00001-enarx-sprint-framework/README.md
npmccallum: As I noted on the call today, I think the right way to move forward here is to treat issues and PRs differently. Sprint planning decides which issues are placed on the sprint board. Issues may or may not have a corresponding PR. All PRs go on the sprint board, regardless of whether they have a corresponding issue or not.

This has the following properties.

1. We prioritize the work of the core team.
2. We can have non-code issues as part of planning. For example, infrastructure issues. These don't need PRs.
3. All PRs have the same process visibility. Whether your work is "core team" work or "outside contributor work" you get the same visibility.
4. Joining the "core team" means you take part in sprint planning. This is open to anyone.

concepts/00001-enarx-sprint-framework/README.md
npmccallum: I think work estimation is indeed important. However, I think it is difficult and I'd like for us to ease our way into this process. What do people think about leaving the unit undefined for now so that we can test our workflow and automation? Then we can add in "work units" and more careful planning as a follow-up.

I'm thinking perhaps we outline the workflow first and then implement it. This puts the RFC in the `demonstrated` state. Then we add "work units" before the transition to `adopted`. Thoughts?

concepts/00001-enarx-sprint-framework/README.md
npmccallum: I agree we need more time. I'd also like planning and retrospective to be back-to-back.

concepts/00001-enarx-sprint-framework/README.md
npmccallum: Agreed. I want roles clearly defined.

concepts/00001-enarx-sprint-framework/README.md
npmccallum: I forgot to mention three things regarding issues.

1. We probably need some sort of issue triage that is more frequent than the Sprint duration. At the very least we want a prompt human touch on each issue that is filed so that contributors feel welcomed. We also need to identify urgent issues...

2. We need to have a notion of "interrupt" issues which bypass sprint planning and go directly to the board even in the middle of a sprint. These should be rare. They should either (1) block existing sprint work or (2) have otherwise catastrophic impact (think: website is down).

3. What is the process for mid-sprint decomposing? What if we have planned an issue for a sprint and while doing it we notice it should be decomposed further? How do we handle this case?

lkatalin: I second that mid-sprint decomposing or some type of check-in re: the chosen work would be beneficial.

concepts/00001-enarx-sprint-framework/README.md
lkatalin: I like the idea of leaving the "work unit" undefined at first and getting the rest of the process working, since I think estimating work will be a challenge in itself.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: An alternative, of course, is to have shorter sprints.  I'm not sure the benefits are worth it, however, but wanted to raise the option.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: > 3. What is the process for mid-sprint decomposing? What if we have planned an issue for a sprint and while doing it we notice it should be decomposed further? How do we handle this case?

Someone needs to make a call (or a specified group of people, or a process for voting, etc.).  This is another reason that we need to define roles and responsibilities.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: @axelsimon - I want to see roles and responsibilities noted.  Sometimes voting isn't appropriate, and the project might need guidance to hit strategic goals, for instance, which might prioritise particular sprints.  One might consider, for instance, a spring dedicated to docs or RFCs which might never make a popular vote, but which are vital to the project.  Product owners have a role here.

concepts/00001-enarx-sprint-framework/README.md
MikeCamel: As recently noted, what "the team" means here is becoming interesting.

concepts/00001-enarx-sprint-framework/README.md
mbestavros: What I was trying to express here was that those discussions around blockers (which are valuable, absolutely) should happen in the context of the issues themselves in the current sprint as opposed to just a quick status update.

If the current sprint board has everything we're working on anyway, we'd be duplicating status by talking about it in an individual report and then later on when we're looking at the board. That's redundant and a potential waste of time, IMO.

Of course, it's not hard at all to add personal reports back in. Just wanted to share some of my rationale.

concepts/00001-enarx-sprint-framework/README.md
mbestavros: @npmccallum My thinking around point 1) of yours above was that we'd do new issue triage _every standup_. This would mean new issues get a human response within 24 working hours. Were you thinking of something in addition to that?

npmccallum: No.

concepts/00001-enarx-sprint-framework/README.md
mbestavros: Resolving to simplify older comments.

concepts/00001-enarx-sprint-framework/README.md
mbestavros: Done in latest version. Resolving to simplify open feedback.

connorkuehl (dismissed):
Looks great!

The images don't appear when viewing the rendered markdown as part of this pull request, but when I navigate to @mbestavros's tree they do render. Not sure if this is a GitHub weirdness or not.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+
+![Complete issue workflow diagram](./full-process.svg)
+
+This flowchart introduces the three crucial events of the process, each of
+which is encapsulated by its own team meeting:
+ 
+ - **Planning**, a meeting representing the start of a new sprint
+ - **Daily Standup**, the team's daily meeting where, among other things,
+   **Triage** is performed.
+ - **Retrospective**, or "Retro" for short, a meeting representing the
```
MikeCamel: Suggest change to "Review" - as discussed in the call, the retro is only part of the review.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+
+## Common Concepts
+
+A few things will be referenced throughout the proposal; they will be defined
+here.
+
+### Triage
+
+Triage is where the team will categorize newly-filed issues and pull requests,
+and where stakeholders will be able to express needed urgency over those items.
```
MikeCamel: Stakeholders are not defined.  This is important, as they are the ones making decisions about what to accept.  If we mean "Product Owner", then let's say so.  If others are involved, we should specify.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+of thumb to follow should be that every team member nominates twice the work
+they think they are able to do in a given sprint.
+
+Once nomination is complete, the **Proposed** board column will be populated
+with a full set of proposed issues to work on for the next sprint. Issues may
+also arrive in this column if they were not started in the previous sprint. The
+process for determining such issues is further detailed in the Retrospective
+meeting (detailed more below).
+
+With that done, the project's stakeholders now have the responsibility to
```
MikeCamel: As noted above, the "stakeholders" are not defined.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+Finally, the rest of the meeting should be dedicated to open-ended project
+discussion. If there are guests to the call, the team should welcome them and
+allow them to ask questions. If there are any other topics to cover, they should
+be brought up here.
+
+Standup should take no longer than a half-hour, and should be scheduled as
+follows:
+
+- 10-15 min: triage
+- 15-20 min: Board review/project discussion (however long is left after triage)
```
MikeCamel: Where does discussion of "blockers" come in?  This is an important part of stand-ups.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+Beyond that, the rest of the meeting should be dedicated to demos or other
+high-level discussion about work accomplished.
+
+Retrospectives should take a full hour. As detailed in [Planning](#planning),
+the two meetings should be combined, with Retrospective happening first,
+replacing the standup of the first day of a new sprint.
+
+Here's a proposed schedule:
+
+- 20 min: review issues/PRs from last sprint
```
MikeCamel: Demos first, if for no other reason that we will sometimes invite visitors, and they shouldn't have to sit through the rest of the meeting.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+- **Issues** on Github take the form of a simple text note in Markdown. Each one
+  has a title and can be assigned metadata such as labels and assignees.
+  
+  They represent work that needs to be planned, implemented, or otherwise
+  completed. An issue may encapsulate any task related to the project, and
+  scope can vary wildly from one issue to the next; where one may be a simple
+  formatting fix, others might be a radical design change. Presently (and for
+  the foreseeable future), the Enarx project uses them exclusively to organize
+  tasks yet to be done, discussion topics, and more (even when not strictly
+  needed).
```
MikeCamel: To be accepted, an issue **must** be well-scoped, and have a definition of "done".  Those writing issues are encouraged to specify or suggests tests to help the assignees to complete the task as expected.

Each issue should also be small enough to be accomplished within one Sprint: if not, it should be decomposed.  In the rare situations where this is not possible, sign-off by the Product Owner is required to spread it over multiple Sprints.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+now more feasible for the Enarx project to begin enforcing stricter merge
+requirements.
+
+A pull request should only be merged when:
+
+- it has two or more approving reviews
+- every reviewer that has submitted a review has updated their review to
+  "Approved"
+
+The latter exists to ensure that all feedback from all reviewers is addressed
```
MikeCamel: Do we want to require tests for code-based PRs?

MikeCamel (changes requested):
My key request is better definition of the approval process, whether that's through disambiguation of "stakeholders" or further description.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+for good examples. And, of course, Red Hat
+[itself](https://developers.redhat.com/blog/2017/06/29/agile-software-development-the-red-hat-way/)
+uses it.
+
+## Unresolved questions
+
+The most significant unresolved question does not involve process, but rather
+the automation we hope to pair with it: whether the project board will
+live inside the main Enarx repository, or as an all-encompassing Enarx
+organization board.
```
lkatalin: Isn't this question resolved now? I see under "Project Board" it claims that the board will be organization-level.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+  the criteria for merge (detailed [below](#merge-criteria)).
+- If the review requests changes, it moves into the `reviewed` state. A PR can
+  arrive here either through an assigned reviewer submitting a review, or
+  through a "drive-by" reviewer that requests changes.
+
+Once a PR is in the `reviewed` state, automation will take care of updating
+assignees to reflect whose responsibility the PR is at the present time:
+
+- The reviewer that submitted the review is unassigned.
+- The PR author is assigned to the PR to address the reviewer's feedback.
```
lkatalin: Maybe this is implied, but it seems that the PR author should only be re-assigned to the PR if the reviewer did in fact request changes. Another option is the reviewer approved, in which case the re-assignment of the PR author is unnecessary. I think it would be annoying for a PR author to see they are re-assigned to an issue, only to check it and see they have no changes to make.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+
+### Workflow
+
+Here's a flowchart of the new and improved proposed workflow for a pull request,
+starting with a PR's submission and going all the way to its merge:
+
+![PR workflow diagram](./pr-workflow.svg)
+
+As with every issue, every PR will be reviewed through a Triage, during which
+reviewers will be requested and assigned. At least two reviewers must be
```
lkatalin: Does this mean that if a PR comes in and you feel motivated to review it, you should *not* review it until the next day's triage? Or does it mean that you can review it and become and "optional reviewer", and other reviewers may be assigned the next day? The latter makes more sense to me.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+
+### Planning
+
+Sprint planning constitutes the beginning of a new sprint. It's where the team
+and project leadership collectively decide what the project's focus should be.
+
+The sprint planning process is illustrated in the following flowchart:
+
+![Sprint Planning Diagram](./sprint-planning.svg)
+
```
lkatalin: This is a detail, but what do the diamonds vs. squares mean?

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+The meeting will start with a triage, as described above. This exists to ensure
+no issues are uncategorized before the planning step. (This is necessary since
+sprint planning will supersede the day's standup and, consequently, its triage,
+as described later.)
+
+With the backlog fully up to date after the triage, the entire team will take
+a small amount of time to **nominate** issues from the area they work on that
+they think are most compelling. It should be each individual team member's
+responsibility to nominate issues of interest _before_ planning begins;
```
lkatalin: This is a bit buried, but I think is very important so that we don't get bogged down in going through every single backlog issue during nominations or have to wait while people make decisions. An inefficient process here would cut into our time for targeted discussion or make it likely that the meeting runs over-time.

In fact, I would argue that during triage, we should have the choice of moving non-urgent issues either to backlog or, if likely to be a priority for next sprint, then straight to "nominated"; this allows us to speed up our individual pre-planning nomination process by being able to quickly throw a small amount of issues from "nominated" into backlog if we want to de-prioritize them, rather than combing through a large amount of backlog issues to remember what we wanted to add to "nominated." 

If the above idea is too unorthodox for sprints, at the very least I think the responsibility described here that the development team moves issues into "nominated" before planning should be a semi-hard requirement. Last minute-nominations can make sense, but they should be the exception and the time limit on this should be enforced.

This way we have adequate time to discuss all the nominations and we minimize (individually and as a team) the process of searching through dozens of backlog issues to find a specific issue we know we want to nominate (but may have forgotten the exact wording of the title, etc.) while preparing for sprints.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+Thus, it will be sent to the next Planning meeting as a Proposed issue (see
+[Planning](#planning) for more details).
+
+If it _does_ have an associated PR, the next question to ask is whether that
+associated PR was merged or not. If it was, then the issue is completed. If not,
+that means the PR is still in review and likely just needs a bit more review
+time. Thus, it should stay as-is in the Under Review column.
+
+Beyond that, the rest of the meeting should be dedicated to demos or other
+high-level discussion about work accomplished.
```
lkatalin: Should demos be assigned as part of the sprint? Does the sprint take into account the time spent on creating a demo in addition to resolving issues? Do we care if we end up with zero demos or a very large number of demos for a sprint? Do we want to have bounds on this?

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,542 @@
...
+
+The decision to make triage a part of daily standup is deliberate: it allows
+issues and PRs filed by "drive-by" contributors (that is, people not part of
+the project that still wish to contribute) to receive human consideration after
+a maximum of 24 working hours -- a desirable property for open source.
+
+Following triage, the scrum master will pull up the sprint board for an informal
+board review. This does not need to go over every single issue; rather, it
+should be an opportunity for team members to point out highlights or interesting
+updates to their tasks, and request help on blockers or otherwise.
```
lkatalin: I think it would be helpful to have a "show of hands" for this part of the meeting re: how many people want to talk about a highlight or blocker. The "round the horn" process we use currently has the advantage of making it immediately obvious how many people will want to say something (i.e., the total number in the meeting) and so pacing of the meeting flows from this naturally. If an unknown number of people have comments to make during this portion, we run into the danger of either 1) the meeting being cut short before everyone gets their say, or 2) the meeting running over-time. Therefore I propose a quick 5-second check before we move into this portion of how many people will bring up a discussion item, like a mini-agenda.

lkatalin (changes requested):
This generally looks really good to me and I think the diagrams are very helpful!
I had a few comments and questions inline.

One additional thought I had (not inline) is: are the sprint boards archived in some way so that we can easily look back on what we prioritized and accomplished during previous sprints? I think this could be very useful both to assess our own work and show what we have done to others.

00001-enarx-sprint-framework/README.md
mbestavros: Yeah, it might be helpful to update the wording here. The "reviewed" state really indicates "changes requested" I think.

I'll also note there is a separate state for explicit approval, and that doesn't involve any reassignment. It's true, though, that this doesn't handle a "review that doesn't request changes". In either case, the submitter doesn't need to do anything.

00001-enarx-sprint-framework/README.md
mbestavros: Yes, the latter.

The implication for someone submitting a "changes requested" review is that they're on board to keep reviewing the PR until their feedback is resolved. That shouldn't need to wait for a formal triage process. That's mostly there to ensure that PRs _have_ the necessary reviewers.

00001-enarx-sprint-framework/README.md
mbestavros: You are right. I've updated it locally and will push it up soon.

00001-enarx-sprint-framework/README.md
mbestavros: I personally like that idea. Maybe it can be a new state in the triage process flowchart if the issue is decided to be "not urgent".

00001-enarx-sprint-framework/README.md
mbestavros: Great question.

For the moment, I'd imagine we'd have to lump the cycles needed to produce a demo in with the overall task, or define making the demo as a separate task. In any case, this is something the team would have to discuss during the prioritization/bandwidth discussion part of sprint planning.

00001-enarx-sprint-framework/README.md
mbestavros: They're standard (ish) ISO flowchart shapes: https://en.wikipedia.org/wiki/Flowchart

I've added an appendix that explains what they mean in the context of this RFC. Pushing soon.

00001-enarx-sprint-framework/README.md
mbestavros: Not a bad idea! I've added that in my latest revision.

mbestavros:
My latest push addressed @lkatalin's review.

> One additional thought I had (not inline) is: are the sprint boards archived in some way so that we can easily look back on what we prioritized and accomplished during previous sprints? I think this could be very useful both to assess our own work and show what we have done to others.

Given we use the same sprint board for every sprint, it'd be hard to preserve what _exactly_ was accomplished in a given sprint including states, etc. If we _really_ wanted to have that, we could maybe look at having a milestone for every sprint, but I'm not sure that's ideal.

For now, the best answer to what we've done is "what's been closed/merged".

00001-enarx-sprint-framework/README.md
mbestavros: I suppose I assumed that was implied. Code tests are currently required to pass anyway. We can make that an explicit requirement for merge if desired. (Makes more sense thinking about it, especially in the context of potentially auto-merging some PRs.)

00001-enarx-sprint-framework/README.md
mbestavros: During board review, as detailed a paragraph or two back. I've put in a small reminder on that line in my next push to make it clear, though.

00001-enarx-sprint-framework/README.md
mbestavros: Agreed, good point. Fixed in my next push.

00001-enarx-sprint-framework/README.md
mbestavros: Sure, makes sense. I've updated to reflect the name change in my next push.

00001-enarx-sprint-framework/README.md
mbestavros: Fixed in latest push.

00001-enarx-sprint-framework/README.md
mbestavros: Fixed in latest push.

00001-enarx-sprint-framework/README.md
mbestavros: I've added a section under "Planning" called "Acceptance Criteria" for this in my latest push.

mbestavros:
My latest push addresses @MikeCamel's feedback, specifically:

- refactored "Retrospective" to be "Review" (where the Retrospective process is a component);
- replaced "stakeholders" with "product owners" to be explicitly clear who those people are;
- added subheaders to Planning, and added a section on what an issue needs to be accepted.

I've also updated several of the diagrams with a better key and updated naming in some cases.

npmccallum:
@axelsimon @blazebissar @m-dombrowski @lkatalin @connorkuehl @MikeCamel 

I believe we have addressed all your comments and are nearly ready to merge. Can you please mark your comments as resolved where appropriate so that we can identify which issues still need to be addressed? Thank you!

00001-enarx-sprint-framework/README.md
lkatalin: I didn't see where / if this was addressed, but maybe since it's a small detail it makes more sense to discuss as a team rather than put it in the document. Especially as I don't know others' thoughts on it.

lkatalin (dismissed):
Thanks for addressing my questions and comments.

jyotsna-penumaka (dismissed):
This looks good to me.
Good work Mark!

haraldh (dismissed):
Looks good. Let's see how it works out in practice.

00001-enarx-sprint-framework/README.md
axelsimon: > Should demos be assigned as part of the sprint? Does the sprint take into account the time spent on creating a demo in addition to resolving issues?

Normally, they should. Demos should really be just that, a demonstration of what has been accomplished during the sprint, so the whole team and stakeholders can come around what they have built and see progress. This is based on the idea that each sprint should deliver incremental value, and that every sprint as such always produces working code. We may be still a bit away from that, so i would loosen that requirement to just demonstrating what has been worked on and progress there, not adamantly requiring that it is a "usable increment".

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+sprint. It is a time for the whole team to reflect on work accomplished, demo
+new things, and discuss how to approach tasks that didn't get completed.
+
+Most of these topics will take the form of discussion. The only significant
+process to detail here is **Retrospective**, the workflow for reviewing issues
+and PRs from the last sprint.
+
+![Sprint Retro diagram](./sprint-retro.svg)
+
+Any issues that don't have an associated PR are assumed to be "not started".
```
axelsimon: This assumes that issues are not code-related (outreach, documentation, mentoring) are never started. That being said, this can be manually sorted out, but there is some bias towards code-related issues here.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+
+### Review
+
+The final meeting is **Review**, which represents the end of a
+sprint. It is a time for the whole team to reflect on work accomplished, demo
+new things, and discuss how to approach tasks that didn't get completed.
+
+Most of these topics will take the form of discussion. The only significant
+process to detail here is **Retrospective**, the workflow for reviewing issues
+and PRs from the last sprint.
```
axelsimon: I think calling this meeting **Review** is a good idea, however i take issue with calling **Retrospective** something that is disconnected from the [accepted meaning](http://retrospectivewiki.org/index.php?title=Agile_Retrospective_Resource_Wiki) of an agile retrospective, especially if we are telling folks we are basing our process on agile processes.

In light of this,
1. we should actually make time for engaging in retrospectives, as they are a good tool to ensure the "agile scaffolding" so to speak is working out for the team members and
2. call what is currently being termed "retrospective" an issue review, a card review, an end of sprint triage, or something like that.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+ - **Daily Standup**, the team's daily meeting where, among other things,
+   **Triage** is performed.
+ - **Review**, a meeting representing the end of a sprint where, among other
+   things, **Retrospective** is performed.
+
+Once an issue is filed, it will be categorized and discussed in Triage. If
+urgent, it's injected straight into the current sprint; if not, it gets pushed
+to the backlog. From there, the rest of the cycle will carry it through to
+completion: Planning will sooner or later land it as part of a sprint, where it
+will become a team member's action item to complete. At the end of the sprint,
```
axelsimon: > Planning will sooner or later land it as part of a sprint

Probably, but there is no guarantee of this: what happens if an issue keeps staying in the backlog?

Not needed right now, but we should probably eventually add a time to cull old issues that are not getting put in sprints, as they have become stale. If they are in fact not stale, why they are not getting added to sprints should in itself be a question worth answering, as it's a sign of some other problem.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+[Planning](#planning) for more details).
+
+If it _does_ have an associated PR, the next question to ask is whether that
+associated PR was merged or not. If it was, then the issue is completed. If not,
+that means the PR is still in review and likely just needs a bit more review
+time. Thus, it should stay as-is in the Under Review column.
+
+Retrospective should happen immediately following demos (if there are any).
+Putting demos first allows those interested to see them without having to sit
+through the minutiae of project discussion.
```
axelsimon: This is good and makes sense, but makes even more sense if you consider/make explicit that all stakeholders should be invited to the demo: some of the stakeholders have no interest in the minutiae of project discussion but are very interested in seeing the progress made. We start with the demo to let them leave after.

Suggestion, add somewhere above the following:

> All stakeholders are invited to the demo, so they can get a sense of the progress made in the last sprint, or since they were last able to attend a demo.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+With a clear framework in place for reviewing and approving pull requests, it's
+now more feasible for the Enarx project to begin enforcing stricter merge
+requirements.
+
+A pull request should only be merged when:
+
+- all required code tests pass
+- it has two or more approving reviews
+- every reviewer that has submitted a review has updated their review to
+  "Approved"
```
axelsimon: All of these are good. However, while i realise these checks are more on the "tickbox" level and this may not fit here, there is no mention of documentation, neither here nor, to the best my reading, in the overall document.

We could possibly add something as simple as:

```suggestion
- all required code tests pass
- it has two or more approving reviews
- every reviewer that has submitted a review has updated their review to
  "Approved"
- any changes that require documentation to be understandable and/or maintainable have the required associated documentation
```

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+
+With that done, the product owners now have the responsibility to prioritize
+the proposed work. If it's something that should get done in the next sprint,
+the stakeholders can move it to **Accepted**; if it can wait, it should be
+moved back to the backlog.
+
+#### Acceptance criteria
+
+It is the product owner's responsibility to only accept issues that meet
+acceptance criteria::
```
axelsimon:
```suggestion
acceptance criteria:
```

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+
+- it must be well-scoped, and have a clear definition of completion
```
axelsimon: Cf. comment on line 420 and following.

```suggestion
- it must be well-scoped, and have a clear definition of completion, including documentation if needed
```

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+
+If an issue cannot be accomplished in a single sprint, it should be decomposed.
+In the rare scenario that an issue cannot be decomposed, it needs to be signed
+off by the product owners to be included in a sprint.
+
+Additionally, the product owners must be careful to accept a reasonable amount
+of work from Proposed and not overload the team members. The purpose of this
+step is not to express that all work is important; that's counterproductive and
+doesn't accomplish anything.
+
```
axelsimon: I really like both of these.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,593 @@
...
+  process. The modular design of our board supports that -- for example, if we
+  consider a new issue or pull request particularly important, we can manually
+  assign it and have it skip the backlog entirely.
+- Moving to a single, organization-level board allows visitors to see a snapshot
+  of what's currently a top priority for Enarx.
+- With automation, this process has potential to minimize the busywork that
+  project developers have to deal with, and only require human judgment when it
+  is needed.
+- Running through a triage every day ensures issues are promptly reviewed by the
+  team after they are submitted.
```
axelsimon: I have not seen been done before, but i like the idea. I'm looking forward to seeing how this works out, as i think the idea of spreading out a backlog grooming, doing a bit every day, is quite smart.

axelsimon (changes requested):
Overall, very neat! Great work @mbestavros and @npmccallum.

A few small things, a general-level concern about documentation and one gripe about the term retrospective, which i really think we should not use for what we want to describe here, but otherwise looks good to go.

00001-enarx-sprint-framework/README.md
mbestavros: The natural fix here is "Any issues that don't have an associated PR **and are still open**." Though you're still right that this doesn't communicate that "work has been done, but it's not quite shippable yet" -- though, then again, that's also a problem with issues that might be close to a PR, but not there yet.

The key thing here is a deliverable. For code issues, that's a PR. For non-code issues, that could be any number of other things. I can specify to make that a little more clear.

00001-enarx-sprint-framework/README.md
mbestavros: Good point. I will add something to the Future Possibilities section.

00001-enarx-sprint-framework/README.md
mbestavros: While perhaps not explicitly stated, I intended for point 1) to be addressed in the last 20 minutes of the meeting ("demos or other discussion" -- which should be simplified to "other discussion".)

I will clarify that those last 20 minutes should be for "retrospective".

As for 2)... I'm not really a fan of conflating "triage" with an "end of sprint triage". What about something like "end of sprint issue cleanup"?

00001-enarx-sprint-framework/README.md
mbestavros: Good point. I've addressed this in my latest version.

00001-enarx-sprint-framework/README.md
mbestavros: Added in my latest revision.

mbestavros:
I've addressed most of @axelsimon's review, though I haven't yet refactored "Retrospective" to be something new.

00001-enarx-sprint-framework/README.md
mbestavros: Fixed, thanks!

00001-enarx-sprint-framework/README.md
mbestavros: Added something to this effect.

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,612 @@
...
+contains the entire project backlog and all current work.
+
+The board will be organized with the following columns:
+
+- backlog: collection of issues not currently being worked on
+- nominated: issues nominated by team members for the next sprint (planning)
+- accepted: issues the team should work on next sprint (planning)
+- assigned: issues actively being worked on by an assigned team member
+- under review: all submitted PRs go here while awaiting review
+- done: all issues and PRs go here when done or merged, respectively
```
npmccallum: We need one additional board: `triage`. The bot should put all open issues that are not in the project board into the `triage` column.

00001-enarx-sprint-framework/README.md
npmccallum:
```suggestion
- all required code tests pass
- it has two or more reviews in the `APPROVED` state
- it has no reviews in the `CHANGES_REQUESTED` state
- it has no reviews in the `DISMISSED` state
```

npmccallum (changes requested):

npmccallum (dismissed):

00001-enarx-sprint-framework/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+## Motivation
+
+The Enarx project needs to organize its work better, and in a more sustainable
+way than before. We need to do a better job planning for medium-term work and
+goals, and in a way that scales more effectively than our prior process.
+
+### Tracking Work
+
+As a Github-native project, Enarx categorizes its work with the two major
```
ueno: Sorry for nitpicking, but I would suggest spelling it as "GitHub".

mbestavros: Fixed!

lkatalin (dismissed):

axelsimon (dismissed):

ueno (dismissed):

haraldh (dismissed):

lkatalin (approved):

MikeCamel (approved):
I can't see a quick way to review issues that I raised, but I believe that they have been, so approving.

npmccallum (approved):

PR #29, opened 2020-05-19
Draft WebAssembly workload format

### LINKED ISSUE #17: Description of workload packaging format required (RFC)
**Description**
We expect to provision more than just a WebAssembly binary to a Keep: data and other binaries may be required.

ueno:
This adds an initial draft of WebAssembly workload format specification.

This is intended for discussion; I am even leaning toward the alternatives while writing :-)
Either way we decide I can try to implement it on top of enarx/enarx#502.

Fixes #17.

### DIFF
--- a/00001-workload-package-format/README.md
+++ b/00001-workload-package-format/README.md
@@ -1,0 +1,96 @@
+# workload-package-format
+- Authors: Daiki Ueno
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-05-19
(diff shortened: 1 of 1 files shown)


MikeCamel (dismissed):
I like this.  Possible further work includes:
1) access to encrypted data (where does the key come from?)
2) access to networked data 
3) access to local storage data.

MikeCamel (approved):
Good description - agree that we should keep looking at the options.

PR #31, opened 2020-05-21
Remove the category directory

npmccallum:
All RFCs are now top level. The tests for name formating have been updated
for the new location. Likewise, documentation has been updated to correct
some incorrect process and improve clarity.


MikeCamel (approved):
I've not checked that the code does what's expected, but I approve the changes to the process doc.

npmccallum:
@MikeCamel I tested the code. Also, the code executed as part of this PR's tests (which passed). So the only potential problem left should be false positive tests.

